PR #30, opened 2014-08-19
Allow to pass parameters to Alice from configuration

sroze:
Related to the pull request made on Alice library (nelmio/alice#130), this pull request allow to add parameters directly from the bundle configuration.

If you look at the example wrote on the Alice's pull request, with this feature here, we simply have to set this configuration in `app/config/config.yml` and remove the `parameters` in fixtures files:
```yml
hautelook_alice:
    parameters:
        medias_path: %kernel.root_dir%/Resources/medias
```


sroze:
@baldurrensch FYI that was merged a long time ago :smiley:

sroze:
@theofidry ?! I was speaking about nelmio/alice#130 that is merged since a lot of time, so this one can be too.

theofidry:
Erf, my bad. Could you debase te PR please?

sroze:
I won't, too much changes, this one should have been merged before, as well as the small others btw.

theofidry:
Yeah, maybe I should, but if possible I would like to avoid as much as possible the legacy work even though I certainly will have to do it at some point. I'll try to add this feature nonetheless, it's an interesting one.

PR #39, opened 2014-10-29
Add command to generate fixtures from table data

baldurrensch:


theofidry:
I think tool like https://github.com/trappar/AliceGenerator are a more appropriate place rather than AliceBundle. As neither you and I don't plan to work on it anytime soon I would rather close it for now.

PR #41, opened 2014-10-30
Inherited DataFixtureLoader from AbstractFixture

harentius:
This changes allow using references (getReference() method) inside alice fixture .yml file. It's convenient when use both Doctrine and Alice fixtures/


theofidry:
@harentius not sure of your use case, but should be good in the new version.

theofidry:
Actually there is a Doctrine loader for it: https://github.com/hautelook/AliceBundle/blob/1.x/src/Doctrine/DataFixtures/AbstractLoader.php. So this can be used instead :)

PR #52, opened 2015-06-18
Fixes #45 - ValueObjects will not be persisted

### LINKED ISSUE #45: Problems with non entity objects and phpcr-odm
currently phpcr-odm throws an exception, when asking for metadata of a non mapped class, so the check in hasIdentity in https://github.com/hautelook/AliceBundle/blob/master/Alice/Loader.php#L187 fails with an exception. Also detaching the references in https://github.com/hautelook/AliceBundle/blob/master/Alice/Loader.php#L109 should honour non entity objects.

I propose to add a method contains to class Doctrine and use that for checking for non entity objects (can do the PR ...)

veewee:
This PR will fix #45 .


veewee:
When are you guys scheduling a new release of AliceBundle?
I would like to get this feature merged.

theofidry:
@veewee could you rebase your work?

veewee:
Hello @theofidry,
It seems that the value objects work in the latest released version.
The codebase has changed too much to rebase, so I think this ticket can be closed.

theofidry:
@veewee glad to hear it. Sorry for so much change in the code base, the upgrade to nelmio/alice 2.0 changed a lot of things.

PR #70, opened 2015-08-22
Added list fixtures command

theofidry:
Added the command `hautelook_alice:fixtures:list` that list all registered fixtures. Inspired from doctrine/DoctrineFixturesBundle#106.

Todo:
* [ ] Add tests
* [ ] Update documentation `h:f:l` will no longer be available as it will be ambigus

@hjr3 could you give me your opinion regarding this feature? I'm not sure of the implementation either, this could be an option `--list` of the current command rather than a new command.

### DIFF
--- a/Doctrine/Command/ListFixturesCommand.php
+++ b/Doctrine/Command/ListFixturesCommand.php
@@ -1,0 +1,96 @@
+<?php
+
+/*
+ * This file is part of the Hautelook\AliceBundle package.
+ *
+ * (c) Baldur Rensch <brensch@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Hautelook\AliceBundle\Doctrine\Command;
+
+use Doctrine\Common\DataFixtures\Purger\ORMPurger;
+use Doctrine\Common\Persistence\ManagerRegistry;
+use Hautelook\AliceBundle\Alice\DataFixtures\LoaderInterface;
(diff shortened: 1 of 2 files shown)


Doctrine/Command/ListFixturesCommand.php
```diff
@@ -0,0 +1,96 @@
...
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Helper\DialogHelper;
+use Symfony\Component\Console\Helper\QuestionHelper;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Question\ConfirmationQuestion;
+
+/**
+ * Command used to list the fixtures.
```
Soullivaneuh: We can simply put "List registered fixtures." IMO.

Doctrine/Command/ListFixturesCommand.php
```diff
@@ -0,0 +1,96 @@
...
+                'Bundles where fixtures should be looked in'
+            )
+        ;
+        //TODO: set help
+    }
+
+    /**
+     * {@inheritdoc}
+     *
+     * \RuntimeException Unsupported Application type
```
Soullivaneuh: `@throws`

theofidry:
If someone wants to finish it, feel free to give it a shot, I would rather add to the new version instead to be honest.

PR #71, opened 2015-08-22
Made fixtures path configurable

theofidry:
Suggested by doctrine/DoctrineFixturesBundle#139.

* [ ] Tests
* [ ] Doc

### DIFF
--- a/Finder/Finder.php
+++ b/Finder/Finder.php
@@ -24,7 +24,23 @@ use Symfony\Component\HttpKernel\KernelInterface;
  */
 class Finder
 {
-    const BUNDLE_FIXTURES_PATH = 'DataFixtures/ORM';
+    /**
+     * @var string
+     */
+    private $bundleFixturesPath;
+
+    /**
+     * @example
+     *  new Finder('DataFixtures/ORM')
+     *  new Finder('Resources/fixtures/ORM')
+     *
+     * @param string $bundleFixturesPath
+     */
+    public function __construct($bundleFixturesPath)
+    {
+        $this->bundleFixturesPath = $bundleFixturesPath;
+    }
+
 
     /**
      * Looks at all the bundles registered in the application to return the bundles requested. An exception is thrown
@@ -67,7 +83,9 @@ class Finder
      * @param BundleInterface[] $bundles
      * @param string            $environment
      *
-     * @return string[] Fixtures files real paths.
+     * @return string[]                  Fixtures files real paths.
+     * @throws \InvalidArgumentException Invalid loader path or no fixtures could be found.
+     * @throws \RuntimeException         No valid path founds in which loading fixtures
      */
     public function getFixtures(KernelInterface $kernel, array $bundles, $environment)
     {
@@ -82,7 +100,7 @@ class Finder
 
             $fixtures = array_merge($fixtures, $this->getFixturesFromDirectory($path));
         }
-
+        
         if (0 === count($fixtures)) {
             throw new \InvalidArgumentException(
                 sprintf('Could not find any fixtures to load in: %s', "\n\n- ".implode("\n- ", $loadersPaths))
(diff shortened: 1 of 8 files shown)


jvasseur:
Making the path configurable has the downside of preventing bundles of knowing where to place their fixtures files for them to be loaded.

But i would still prefer a configurable path because the default one doesn't seems logical to me (fixtures are not php files and should be placed somewhere in the Resource directory).

Finder/Finder.php
```diff
@@ -24,7 +24,23 @@
...
+    private $bundleFixturesPath;
+
+    /**
+     * @example
+     *  new Finder('DataFixtures/ORM')
+     *  new Finder('Resources/fixtures/ORM')
+     *
+     * @param string $bundleFixturesPath
+     */
+    public function __construct($bundleFixturesPath)
```
Soullivaneuh: This is BC break if the finder constructor is extended.

Finder/Finder.php
```diff
@@ -82,7 +100,7 @@ public function getFixtures(KernelInterface $kernel, array $bundles, $environmen
 
             $fixtures = array_merge($fixtures, $this->getFixturesFromDirectory($path));
         }
-
+
```
Soullivaneuh: Trailing white-space. :heart:

theofidry:
Thanks for the feedback @jvasseur.

Until now I've never really bothered with the path as I kept the same as DoctrineFixturesBundle. Besides:

* Although Fixtures files are usually YAML files, they can be PHP files as well
* You may have data loaders, which are PHP classes, mixed with your fixtures to have a more granular loading

That being said I'm sensible to your point and seeing that some other fixtures bundle libraries prefer to put that in `Resources`, I wouldn't mind changing this convention.

@Soullivaneuh this PR is a bit old and needs to be rebased, plus as you pointed out they is some BC. I think it's time to make master v2-dev and create branches for v1, I'll take care of that tonight.

theofidry:
Btw @Soullivaneuh should I give you write access? Given your commitment and me being a bit overwhelmed by work recently that would only be fair.

jvasseur:
Data loaders could stay in the DataFixtures namespace while fixtures files are put inside the Resource directory.

The convention for bundles is to keep non-class files in the Resource directory. Even php files that doesn't contain a class can be kept in the resource directory (it is the case for scripts in the SensioDistributionBundle).

Soullivaneuh:
> Btw @Soullivaneuh should I give you write access? Given your commitment and me being a bit overwhelmed by work recently that would only be fair.

I would be glad to be part of contributors team and spend some of my free time on this project when I got some. ;-)

Just e-mail me the "must follow" rules of hautelook projects and we are good to go. :-)

Soullivaneuh:
> I think it's time to make master v2-dev and create branches for v1, I'll take care of that tonight.

Maybe, but this BC break is easy to fix AFAIK.

Just add a default value (`null`) on the constructor and manage it.

theofidry:
@weaverryan would you require this to be merged for your tutorial as well or it's really no big deal?

weaverryan:
No big deal for me - it makes sense though :)

theofidry:
I'll try to see in the week how it goes. But I think it could be added as a feature and tagged as 1.2.0. The only annoying bit is that fixtures should be in the `Resource` folder instead of `DataFixtures`, as are no longer classes unless using loaders (which are a legacy of DoctrineFixturesBundle).

Well, nitpicking I guess :dancer:

mpdude:
Will you allow to change the path via console arguments?

theofidry:
@mpdude is there any point in doing that? For sure it's doable but I just don't really see the need.

Btw this PR is a bit in stalled stated as right now I'm more focusing on nelmio/alice directly, so feel free to take over that PR and finish it :P

mpdude:
My use case was that I have a `fixtures/`directory with various subdirectories for various scenarions/things I'd like to test/demonstrate.

With such a switch, you can run `bin/console ... --fixtures=fixtures/scenario1` and so on to load just the things you need.

theofidry:
Fair. I'm for including it :)

theofidry:
I could not find the time to finish it so if someone wants to work on it... It is already included for AliceBundle 2.x, so it's just that I prefer to focus on other things before considering spending time on backporting this feature. I don't think there is a lot missing, but it's just that I don't have enough time for it :/

Issue #81, opened 2015-08-25
Public method in the Loader - Not Used for Custom Faker Data Providers


kyaw-myo-thein:
**In v0.2**, 
we could use public method in the loader as custom faker data providers as in [documentation](https://github.com/nelmio/alice/blob/master/doc/customizing-data-generation.md#public-method-in-the-loader). As in e.g, we can use ```name: <groupName()>``` to generate specific group names.

**In v1.0.0-beta**, 
I realised that public methods are not recognised as data provider anymore. In my case, providers are useful only in the context of loader class, so I don't wish to [create separate faker class]( 
https://github.com/hautelook/AliceBundle/blob/master/Resources/doc/faker-providers.md)

theofidry:
I could identify the bug and it's a rather nasty issue. So it will be solved shortly but have to wait a bit.

theofidry:
@kyaw-myo-thein this issue should be solved now, I let you confirm it.

kyaw-myo-thein:
I'll try to give feedback tomorrow (Sydney time).

I had a brief look today, checking out master branch. Switching to ```hautelook_alice:fixtures:load``` command throws unexpected error. I need to investigate further.

kyaw-myo-thein:
It looks good now. Thank for fixing it @theofidry 

I'll close this issue.

PR #83, opened 2015-08-25
Loaders as providers

theofidry:
As addressed in issue #81, data loaders are no longer took as providers. As a result, public methods of data loaders are no longer recognised as formatters.

To solve it, the Doctrine Finder now look for data loaders too and a new loader, taking the data loaders found as providers too, will be used in the command.

### DIFF
--- a/Doctrine/DataFixtures/AbstractLoader.php
+++ b/Doctrine/DataFixtures/AbstractLoader.php
@@ -31,6 +31,6 @@ abstract class AbstractLoader extends ContainerAware implements LoaderInterface
      */
     public function load(ObjectManager $objectManager)
     {
-        return $this->container->get('hautelook_alice.fixtures.loader')->load($objectManager, $this->getFixtures());
+        throw new \RuntimeException('This method was not expected to be called.');
     }
 }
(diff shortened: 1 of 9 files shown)


Doctrine/DataFixtures/AbstractLoader.php
```diff
@@ -31,6 +31,6 @@
      */
     public function load(ObjectManager $objectManager)
     {
-        return $this->container->get('hautelook_alice.fixtures.loader')->load($objectManager, $this->getFixtures());
+        return [];
```
theofidry: Might be better to throw an exception instead

Doctrine/Finder/Finder.php
```diff
@@ -47,13 +49,40 @@ public function getFixturesFromDirectory($path)
...
+     * Gets all data loaders instances.
+     *
+     * For first get all the path for where to look for data loaders.
+     *
+     * @param BundleInterface[] $bundles
+     * @param string            $environment
+     *
+     * @return LoaderInterface[] Fixtures files real paths.
+     */
+    public function getDataLoaders(array $bundles, $environment)
```
theofidry: Needs tests

Issue #72, opened 2015-08-23
support for other doctrine flavors


dbu:
there is currently only an ORMExecutor. it would be cool to have executors for the other doctrines

* [ ] PHPCR-ODM executor
* [ ] Mongodb-ODM executor
* [ ] Couchdb-ODM executor

PR #86, opened 2015-08-26
Support other Doctrine flavours

theofidry:
Solves #72.

@dbu I'm not really sure of how I should wrap it up with the command. Should we pass an option to the command? Have several commands?

Doctrine flavours:
  * [x] Doctrine ORM
  * [x] MongoDB
  * [x] PHPCR

Other tasks:
  * Update tests
    * [x] Doctrine ORM
    * [x] MongoDB
    * [ ] PHPCR
  * [x] Add deprecation notice for previous command
  * [x] Update Doc & Upgrade guide


dbu:
in phpcr we namespaced our commands with `doctrine:phpcr`, for example `doctrine:phpcr:mapping:info`. i think its going to be easier to have separate commands per storage layer. with some factory methods, you could reuse most of [LoadDataFixturesCommand](https://github.com/hautelook/AliceBundle/blob/master/Doctrine/Command/LoadDataFixturesCommand.php). (btw, the name of that command seems a bit wrong, as its not saying its about doctrine. but it is specific to doctrine orm...)

i would propose:
* hautelook_alice:doctrine:load (and keep the fixtures:load command registered for BC but deprecate it)
* hautelook_alice:doctrine:phpcr:load
* hautelook_alice:doctrine:...:load (not sure what mongo / couch call their commands)

you could do a compilar pass removes any of the commands services that require a doctrine flavour that is not available. (you will notice when the manager registry service is missing)

dbu:
oh, and great that you work on this!

theofidry:
Hm, didn't though on the compiler pass thing to remove unused things, good to hear.

Well for the command name although there's no "Doctrine" in it's in the Doctrine namespace so it would be kind of redundant IMO.

Regarding CouchDB, I didn't found any reference in doctrine/data-fixtures.

For the PR I'm not sure on when I'll be able to finish it as the next weeks promise to be very busy, but I'll do my best to include it in the 1.0.0 release.

dbu:
would be great to bring it in.

RE command name: i did not mean the command class (and if you move to more injection you might even avoid tying it to  a specific doctrine flavour). the name of the command is a problem, as its called `hautelook_alice:fixtures:load` which does not say doctrine although its specific to doctrine.

lets ignore couch. no idea how active it is...

theofidry:
Indeed, and as we may add a support for Propel in the next versions it would make more sense too.

theofidry:
@dbu ready for review :)

Still have some tests to do though

dbu:
awesome! one detail: please typehint the DocumentManagerInterface for phpcr
to allow for decorated dm's.

apart from that lets merge and make sure the phpcr-odm and cmf docs
reference this bundle in the relevant places.

theofidry:
@dbu somewhat ending up refactoring the `Finder` to create and interface and split it into `FixturesFinder` and `BundlesResolver`.

I've added a `Document/Product` entity for tests with Doctrine ODM and `Document/Task` for tests with Doctrine PHPCR-ODM. But can't get it work with `Task`, I get the following error: 

```
[Doctrine\ODM\PHPCR\Id\IdException]
  ID could not be read from the document instance using the AssignedIdGenerator.
```

Not sure from where the error is coming from, do you have any idea?

dbu:
the default ids are not autogenerated, you need to set it explicitly or set
it to auto generated id. its in the phpcr odm doc, hope you find it. i am
travelling atm and only have my mobile phone with me...

theofidry:
Erg, kind of not working unsure why. But don't worry if you can take a look later it's fine for now I need to update the tests anyway so it's not that urgent.

theofidry:
@dbu there's only how to set this ID thing and then it's good

theofidry:
I'll be merging it for now. There's only the tests for PHPCR that are skipped otherwise is all green.

ReservedDeveloper:
In loading the latest from master, and only the [ODM library](https://github.com/doctrine/mongodb-odm) and [ODM Bundle](https://github.com/doctrine/DoctrineMongoDBBundle), I'm getting the following error on the console:

```sh 
[Symfony\Component\DependencyInjection\Exception\ServiceNotFoundException]                                                                                    
  The service "hautelook_alice.doctrine.generator.loader_generator" has a dependency on a non-existent service "hautelook_alice.doctrine.orm.fixtures_finder".
```

I've got the ODM & Alice bundles enabled in the AppKernel, and the following config block:
```yml
hautelook_alice:
    locale: en_US       # Locale to used for faker; must be a valid Faker locale otherwise will fallback to en_EN
    seed: 1             # A seed to make sure faker generates data consistently across runs, set to null to disable
    persist_once: false # Only persist objects once if multiple files are passed
    db_drivers:
        orm: false #attempting to manually disable ORM to prevent error
        mongodb: true
doctrine_mongodb:
    connections:
        default:
            server: "%mongo_database_host%"
            options: {}
    document_managers:
        default:
            auto_mapping: true
```

Is ORM still a dependency after this merge? Or am I missing a configuration option somewhere?

theofidry:
ORM is definitely not a dependency, but your configuration should be good. Which version are you using? If it's not already the case, try to upgrade to the `@beta.3` and open a ticket if the problem persists.

ReservedDeveloper:
I'm currently using `dev-master`. I'll switch to `@beta.3` and report back with a ticket if the problem doesn't resolve itself.

theofidry:
The beta.3 has just been release so chances are hight that the bug is still
present.

PR #111, opened 2015-09-12
purge-with-truncate option fix

rznzippy:
Previously to this commit, the option "purge-with-truncate" seems didn't worked as InputArgv::hasOption always return true for all defined options even if option is not passed from CLI.

UPD: .idea removed from commit


.gitignore
```diff
@@ -1,3 +1,4 @@
+.idea/*
```
theofidry: you should keep IDE and OS files specific to your global gitignore, not the projects one

theofidry:
Thanks for the patch @rznzippy. There's only the two errors to fix, nevermind the failure I'm fixing that in another PR.

theofidry:
Done, thanks @rznzippy

PR #118, opened 2015-09-14
Fix composer indentation

Soullivaneuh:
Also added an editorconfig file

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,8 @@
+[*]
+end_of_line = lf
+insert_final_newline = true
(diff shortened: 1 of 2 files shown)


theofidry:
Maybe we should set up a StyleCI in place rather than the editorconfig file?

Soullivaneuh:
Both are different.

.editorconfig will ensure most popular IDE to be conform of code styling, that make developer life better. :-) http://editorconfig.org/

StyleCI and PHP-CS-fixer juste make sure dev are not coding so bad. ;-)

Both should be implemented IMO.

PR #119, opened 2015-09-14
Loading limit configuration

Soullivaneuh:
I'm trying to add loading limit on configuration to set a default, but I'm facing on a problem: Your service is never used.

Instead, a instance of loader is created on `LoaderGenerator`: https://github.com/hautelook/AliceBundle/blob/e739300431db9fc9706006bde7b5e8d321c11a7d/src/Doctrine/Generator/LoaderGenerator.php#L51-L55

Two questions:

* Why? :-)
* How to handle this parameter?

### DIFF
--- a/src/Alice/DataFixtures/Loader.php
+++ b/src/Alice/DataFixtures/Loader.php
@@ -42,35 +42,24 @@ class Loader implements LoaderInterface
     /**
      * @var int
      */
-    private $loadingLimit = 5;
+    private $loadingLimit;
 
     /**
      * @param FixturesLoaderInterface $fixturesLoader
      * @param ProcessorInterface[]    $processors
      * @param bool                    $persistOnce
+     * @param int                     $loadingLimit
      */
     public function __construct(
         FixturesLoaderInterface $fixturesLoader,
         array $processors,
-        $persistOnce
+        $persistOnce,
+        $loadingLimit
     ) {
         $this->fixturesLoader = $fixturesLoader;
         $this->processors = $processors;
         $this->persistOnce = $persistOnce;
-    }
-
-    /**
-     * Sets load file limit, which is the maximum number of time the loader will try to load the files passed.
-     *
-     * @param int $loadingLimit
-     *
-     * @return $this
-     */
-    public function setLoadingLimit($loadingLimit)
-    {
         $this->loadingLimit = $loadingLimit;
-
-        return $this;
     }
 
     /**
@@ -130,6 +119,14 @@ class Loader implements LoaderInterface
     public function getPersistOnce()
     {
         return $this->persistOnce;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getLoadingLimit()
+    {
+        return $this->loadingLimit;
     }
 
     private function areAllFixturesLoaded(array $normalizedFixturesFiles)
--- a/src/Doctrine/Generator/LoaderGenerator.php
+++ b/src/Doctrine/Generator/LoaderGenerator.php
@@ -51,7 +51,8 @@ class LoaderGenerator implements LoaderGeneratorInterface
         return new Loader(
             $_fixturesLoader,
             $loader->getProcessors(),
-            $loader->getPersistOnce()
+            $loader->getPersistOnce(),
+            $loader->getLoadingLimit()
         );
     }
 }
--- a/README.md
+++ b/README.md
@@ -75,6 +75,7 @@ hautelook_alice:
     locale: en_US       # Locale to used for faker; must be a valid Faker locale otherwise will fallback to en_EN
     seed: 1             # A seed to make sure faker generates data consistently across runs, set to null to disable
     persist_once: false # Only persist objects once if multiple files are passed
+    loading_limit: 5    # Maximum number of time the loader will try to load the files passed
 ```
 
 Fore more information regarding the locale, refer to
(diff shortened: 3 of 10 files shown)


theofidry:
Erf right.

> Why? :-)

Because of [this](https://github.com/hautelook/AliceBundle/blob/e739300431db9fc9706006bde7b5e8d321c11a7d/src/Doctrine/Generator/LoaderGenerator.php#L49). Doctrine data fixtures must be added as providers during the loading time, but not permanently.

> How to handle this parameter?

The `LoaderGenerator` should take into account this new parameter

Also it's maybe worth considering rising the current default value. I putted 5 arbitrary without much to back it up...

Soullivaneuh:
> I putted 5 arbitrary without much to back it up...

Yeah I see, the setter is never used. :-)

Soullivaneuh:
Should be better like this.

Also I think the loader service is not necessary, am I right?

theofidry:
I would put the loading limit in the loader service, not the loader generator. The loader generator would then just take the loading limit of the loader service.

> Also I think the loader service is not necessary, am I right?

What do you mean?

theofidry:
> Yeah I see, the setter is never used. :-)

Currently no, it was more if you wanted to change it... But you're right putting it in the configuration and then passing it to the constructor is better.

Soullivaneuh:
Test fixed.

> What do you mean?

Nevermind, I said a mistake.

> I would put the loading limit in the loader service, not the loader generator.

I would be glad to do that, but how?

theofidry:
>> I would put the loading limit in the loader service, not the loader generator.

> I would be glad to do that, but how?

By adding it to the constructor [here](https://github.com/hautelook/AliceBundle/blob/master/src/Resources/config/services.xml#L39). That means to add it to the constructor and remove the useless setter

Soullivaneuh:
> By adding it to the constructor here. That means to add it to the constructor and remove the useless setter. It may require to add a getter though

I can do that, but the problem will be the same: The loading limit must be passed on `LoaderGenerator` because a new instance of `Loader` is created instead of service usage.

theofidry:
If you add the `$loadingLimit` to the constructor of `Loader` and add a `getLoaderLimit()` to the `LoaderInterface`, then you can retrieve the loader limit set in [`$loader`](https://github.com/hautelook/AliceBundle/blob/master/src/Doctrine/Generator/LoaderGenerator.php#L41) for [generating the new loader](https://github.com/hautelook/AliceBundle/blob/master/src/Doctrine/Generator/LoaderGenerator.php#L51).

So something like:

```php
public function generate(
        LoaderInterface $loader,
        FixturesLoaderInterface $fixturesLoader,
        array $bundles,
        $environment
    ) {
        $doctrineDataLoaders = $this->fixturesFinder->getDataLoaders($bundles, $environment);
        $_fixturesLoader = clone $fixturesLoader;
        $_fixturesLoader->addProvider($doctrineDataLoaders);
        return new Loader(
            $_fixturesLoader,
            $loader->getProcessors(),
            $loader->getPersistOnce(),
            $loader->getLoadingLimit()
        );
    }
```

src/Alice/DataFixtures/Loader.php
```diff
@@ -48,29 +48,18 @@ class Loader implements LoaderInterface
...
      * @param ProcessorInterface[]    $processors
      * @param bool                    $persistOnce
+     * @param int                     $loadingLimit
      */
     public function __construct(
         FixturesLoaderInterface $fixturesLoader,
         array $processors,
-        $persistOnce
+        $persistOnce,
+        $loadingLimit = 5
```
theofidry: should not have a default value as the default value is given in the bundle configuration

src/Doctrine/Generator/LoaderGenerator.php
```diff
@@ -51,7 +58,8 @@ public function generate(
         return new Loader(
             $_fixturesLoader,
             $loader->getProcessors(),
-            $loader->getPersistOnce()
+            $loader->getPersistOnce(),
+            $this->loadingLimit
```
theofidry: should take the `$loader` loading limit instead, which requires to add a getter in the `LoaderInterface`

Soullivaneuh:
@theofidry should be ok now.

theofidry:
Hm... strange error from Travis

Soullivaneuh:
Which kind?

theofidry:
Could you update the doc too please? (the configuration part)

theofidry:
```
$ vendor/bin/phpunit -c phpunit.xml.dist
PHPUnit 4.8.7 by Sebastian Bergmann and contributors.
Array to string conversion
```

Soullivaneuh:
> Could you update the doc too please? (the configuration part)

Done.

For phpunit, I don't know. Can't reproduce it on my local...

theofidry:
Same here

Soullivaneuh:
I have a proposition to do to improve Travis test and CS. I could make it now and rebase this PR after to test if we have same issue...

theofidry:
> I have a proposition to do to improve Travis test and CS. I could make it now and rebase this PR after to test if we have same issue...

I'm always all ears for improvements

theofidry:
Looks like it passes now

Soullivaneuh:
Indeed. Very strange. Maybe a Travis issue?

Anyway, this can be merge now! :-)

New beta tag would be appreciated.

theofidry:
Tagged

Soullivaneuh:
Thanks! :+1:

PR #120, opened 2015-09-15
Travis improvements

Soullivaneuh:
* PHP 7
* Multiple Symfony versions tests
* Lowest deps
* Composer file cache
* Composer dist install with github oauth
* Global phpunit install to avoid painful deps
* phpunit-bridge: https://packagist.org/packages/symfony/phpunit-bridge
* Symfony >= 2.3 minimum requirement (2.1 and 2.2 are deprecated).

To get it fully working, you need to set `GITHUB_OAUTH_TOKEN` on Travis.

For the token, no special authorization is needed, you can untick all. This is just for authentication.

Please remember to set `GITHUB_OAUTH_TOKEN` as a private variable so it will be not used on PR for security reasons.


Soullivaneuh:
* Deprecations need to be solved: https://travis-ci.org/hautelook/AliceBundle/jobs/80418672#L477
* Seems to not be compatible with SF 2.3: https://travis-ci.org/hautelook/AliceBundle/jobs/80418677#L209. Do you know why?
* Lowest deps need to be fixed: https://travis-ci.org/hautelook/AliceBundle/jobs/80418676#L394

.travis.yml
```diff
@@ -2,39 +2,62 @@ language: php
...
+  directories:
+    - $HOME/.composer/cache/files
 
 before_install:
-  - composer self-update
+  - composer selfupdate
+  - composer config -g github-oauth.github.com $GITHUB_OAUTH_TOKEN
+  - composer global require phpunit/phpunit --no-update
+  - if [ "$SYMFONY_VERSION" != "" ]; then composer require "symfony/symfony:${SYMFONY_VERSION}" --no-update; fi;
   - bash -c 'echo "extension = mongo.so" >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini'
```
theofidry: let's remove these `bash -c` too

.travis.yml
```diff
@@ -2,39 +2,62 @@ language: php
...
 
 install:
   - bash -c 'if ([ "hhvm" != "$TRAVIS_PHP_VERSION" ] && [ "nightly" != "$TRAVIS_PHP_VERSION" ]); then wget https://scrutinizer-ci.com/ocular.phar; fi;'
-  - composer install --prefer-source --no-interaction
+  - composer global update --prefer-dist --no-interaction
+  - composer update --prefer-dist --no-interaction $COMPOSER_FLAGS
 
 script:
-  - vendor/bin/phpunit -c phpunit.xml.dist
+  - make test
```
theofidry: not a big fan of makefiles, and for just the phpunit it's not really worth it :)

composer.json
```diff
@@ -30,13 +30,13 @@
         "doctrine/phpcr-odm": "^1.2",
         "fabpot/php-cs-fixer": "^1.9",
         "jackalope/jackalope-doctrine-dbal": "^1.2",
-        "phpunit/phpunit": "^4.7",
-        "symfony/console": "~2.1",
-        "symfony/framework-bundle": "~2.1",
-        "symfony/yaml": "~2.1",
-        "symfony/validator": "~2.1",
+        "symfony/console": "~2.3",
```
theofidry: should we really put a higher min version here? If it works with old ones it's always fine, by default composer will take the higher available version anyway

Soullivaneuh:
We got a problem.

This class: https://github.com/hautelook/AliceBundle/blob/37ede707ef779caf55b7c26dcf5a8daff73fefaf/tests/KernelTestCase.php

Extends Symfony KernelTestCase.

The problem is this class doesn't exists on SF 2.3.

How should I solve it?

theofidry:
> Deprecations need to be solved: https://travis-ci.org/hautelook/AliceBundle/jobs/80418672#L477

Should be easy to solve, looks like there's just the factory declaration that is wrong

> Seems to not be compatible with SF 2.3: https://travis-ci.org/hautelook/AliceBundle/jobs/80418677#L209. Do you know why?

Looks like it's PsySH lower dep that causes problem

.travis.yml
Soullivaneuh: Who said it will be just for phpunit? :-)

I have another proposition to make after.

theofidry:
There shouldn't be a need for extending the KernelTestCase, I think it's due to a poor configuration at some time there's some way to do it better.

For the 2.3 you have an idea other than removing the dependence to this class?

composer.json
Soullivaneuh: This is not the problem. The problem is some people would install your bundle with LTS SF 2.3 version. We need to test all maintained version for compatibility.

Soullivaneuh:
> For the 2.3 you have an idea other than removing the dependence to this class?

I have a quickly one, but it's ugly: Copy the concerned case from vendor. :-)

I'm already far away.

Soullivaneuh:
So @theofidry, what should I do?

theofidry:
Not much choice:

1. Don't test Symfony 2.3
2. Copy past KernelTestCase
3. Rewrite the test to avoid relying on KernelTestCase and use directly the Kernel

Not sure what's best, but if we want to provide the support for Sf 2.3 I guess the option 2 is the best :/

Soullivaneuh:
I will choose option 2 with a todo to remove this copy paste when Symfony 2.3 support will be dropped.

Soullivaneuh:
@theofidry Removing you bash line produce this error: https://travis-ci.org/hautelook/AliceBundle/jobs/80608506

Another solution to propose?

theofidry:
It doesn't look like the error is related to the bash line but rather the mongodb php extension. Please keep your commits rather than amending for now so that I can more easily see the diffs :) we can always rebase to squash them before merging

Soullivaneuh:
> It doesn't look like the error is related to the bash line but rather the mongodb php extension.

Both of them so. This error appear since I removed this line. Will restore it for the moment.

theofidry:
Oh ok, no I didn't mean to remove completely the line, the just remplacing `bash -c 'my command here;', by `my command`, it does not apply just to this line but the other ones as well

Soullivaneuh:
Oh ok, I misunderstood.

theofidry:
> Deprecations need to be solved: https://travis-ci.org/hautelook/AliceBundle/jobs/80418672#L477

Looks like it comes from PHPCR, not AliceBundle :/

Soullivaneuh:
Got another weird error on lowest deps: https://travis-ci.org/hautelook/AliceBundle/jobs/80614885#L404

Can't understand why this attributes doesn't work, I got dependencies that should handle it:

```bash
sullivan@sweetnexy:~/projects/fork/AliceBundle(travis-improve)$ composer info -i | grep symfony
symfony/config                        v2.2.0  Symfony Config Component
symfony/console                       v2.3.10 Symfony Console Component
symfony/debug                         v2.3.0  Symfony Debug Component
symfony/dependency-injection          v2.2.0  Symfony DependencyInjection Component
symfony/doctrine-bridge               v2.3.0  Symfony Doctrine Bridge
symfony/event-dispatcher              v2.1.0  Symfony EventDispatcher Component
symfony/expression-language           v2.7.0  Symfony ExpressionLanguage Component
symfony/filesystem                    v2.3.0  Symfony Filesystem Component
symfony/finder                        v2.7.0  Symfony Finder Component
symfony/framework-bundle              v2.3.0  Symfony FrameworkBundle
symfony/http-foundation               v2.2.0  Symfony HttpFoundation Component
symfony/http-kernel                   v2.3.0  Symfony HttpKernel Component
symfony/options-resolver              v2.1.0  Symfony OptionsResolver Component
symfony/phpunit-bridge                v2.7.4  Symfony PHPUnit Bridge
symfony/process                       v2.3.0  Symfony Process Component
symfony/routing                       v2.2.0  Symfony Routing Component
symfony/security                      v2.3.0  Symfony Security Component
symfony/stopwatch                     v2.5.0  Symfony Stopwatch Component
symfony/templating                    v2.1.0  Symfony Templating Component
symfony/translation                   v2.3.0  Symfony Translation Component
symfony/validator                     v2.3.0  Symfony Validator Component
symfony/yaml                          v2.3.0  Symfony Yaml Component
```

Soullivaneuh:
Got it. Bad `symfony/dependency-injection` version.

Soullivaneuh:
Ouch, I found some SF 2.3 BC breaks on your code.

Will rewrite the history to start a proper one to separate each fix.

Soullivaneuh:
Should be ok now. :+1:

Soullivaneuh:
> > Deprecations need to be solved: https://travis-ci.org/hautelook/AliceBundle/jobs/80418672#L477

> Looks like it comes from PHPCR, not AliceBundle :/

You are right. Strange that appear on *your* test suites. Will take a look.

Soullivaneuh:
Another issue: https://github.com/hautelook/AliceBundle/blob/37ede707ef779caf55b7c26dcf5a8daff73fefaf/src/Resources/config/services.xml#L31

Expression argument type is not available on 2.3, we can't use it.

What is the alternative?

theofidry:
Erf... Well if I remember right the only places were I've used it is for passing the processors and providers of the `ProcessorChain` and `ProviderChain`. So we may pass the chains rather than arrays (should be easy fortunately). It's a bit less flexible but more secure as well.

It would be better to do this one in another PR though.

Soullivaneuh:
> It's a bit less flexible but more secure as well.

Yeah, but with that, your bundle is **totally imcompatible** with SF 2.3 full stack.

The other solution would be to drop support SF 2.3. Not very happy of this but we have another LTS version out now: 2.7.

> It would be better to do this one in another PR though.

Maybe, but travis will fail on master if we merge it as is. In addition, this bug can hide anothers...

theofidry:
> It's a bit less flexible but more secure as well.

I meant passing the chains instead of using the expressions, which will be compatible with Sf 2.3.

> Maybe, but travis will fail on master if we merge it as is. In addition, this bug can hide anothers...

Will be a big PR but well, I guess no choice then.

Soullivaneuh:
For example, with processor, it seems `ProcessorChain` is just a bridge to passe references instead of array through expression, am I right?

If yes, we can remove this and pass directly processors on fixtures loader services through compiler pass directly.

What do you think?

Soullivaneuh:
> Will be a big PR but well, I guess no choice then.

This is why I rewritten and will rewrite history to avoid single big commit. ;-)

theofidry:
> For example, with processor, it seems ProcessorChain is just a bridge to passe references instead of array through expression, am I right?

Exactly, and opens it to the users, might not be of much use for the ProcessorChain but it can be for the ProviderChain.

> If yes, we can remove this and pass directly processors on fixtures loader services through compiler pass directly.

We can do that too but. Actually I'm not sure of what's better as IMO the ProcessorChain will be rendered completely useless and the ProviderChain won't have much use either.

So it comes down to this question: is it better to inject to the loaders arrays of processors and arrays of providers or to inject the ProcessorChain and ProviderChain?

theofidry:
Actually I think the second one is better: if you want to ship with your own loader you'll just have to inject the chains services rather than doing a compiler pass and looping. And I prefer manipulating immutable objects ;)

But if you have another though on this I'll gladly listen to your arguments :)

Soullivaneuh:
Sounds great, I'll try by passing directly `ProcessorChain` and `ProviderChain` services instead of result.

What about this one? https://github.com/hautelook/AliceBundle/blob/37ede707ef779caf55b7c26dcf5a8daff73fefaf/src/Resources/config/services.xml#L33

theofidry:
Yeah this one... It's for passing all the application parameters to the Loader, this way you can use parameters in this fixtures files cf [fixtures-parameters](https://github.com/hautelook/AliceBundle/blob/master/src/Resources/doc/advanced-usage.md#fixtures-parameters).

This loader comes form Alice loader so I guess we won't escape the compiler pass for this one.

Soullivaneuh:
For parameters, I passed values through extension class in b38fa69.

Please tell me if it's OK for you.

Soullivaneuh:
One of your project that depends of this one has the same issue: https://github.com/theofidry/PsyshBundle/blob/master/src/Resources/config/services.xml#L13

If my solution seems correct, I can provide a PR for this one too.

theofidry:
> One of your project that depends of this one has the same issue: https://github.com/theofidry/PsyshBundle/blob/master/src/Resources/config/services.xml#L13

Very likely^^ I've never bothered with Sf 2.3 before...

> If my solution seems correct, I can provide a PR for this one too.

Gladly

Soullivaneuh:
I just made `ProcessorChain` as a property for loader, check this out.

theofidry:
IMO this PR start to become a little too big, we're mixing two things: enhance Travis configuration and add supports for older version.

Somehow we could keep a PR for only Travis configuration, and put the Symfony 2.3 test in the allowed failures. Then do PR to add support for Symfony 2.3 and make it pass as a mandatory test (no longer an allowed failure one).

Soullivaneuh:
Why not. But could you answer to my questions? Then I will know what to do after that.

theofidry:
Sorry which one?

Soullivaneuh:
* https://github.com/hautelook/AliceBundle/pull/120#commitcomment-13263742
* https://github.com/hautelook/AliceBundle/pull/120#commitcomment-13263779

Soullivaneuh:
@theofidry I don't understand this error: https://travis-ci.org/hautelook/AliceBundle/jobs/80784140#L209

I didn't touch jackalope dep...

theofidry:
Erf not sure, it has caused me more than one problem to get it installed already... Maybe @dbu has an idea on this.

But IMO we should only keep the first 3 commits you've done for this PR and fix the risen issues on another one. They're not trivial and it will be easier to spot the errors

Soullivaneuh:
> But IMO we should only keep the first 3 commits you've done for this PR and fix the risen issues on another one. They're not trivial and it will be easier to spot the errors.

OK I'll do that.

dbu:
@Soullivaneuh just back from holiday. that update conflict indeed looks strange. did you revert something to get rid of it? if you can recreate the problem in a separate PR and ping me, i am happy to help figuring out what is going on.

Soullivaneuh:
@dbu thanks. This seems to be related to composer.json modification.

I'll ping you when I'll work on it, if I got this problem again. :+1:

PR #126, opened 2015-09-17
CS fixer automation

Soullivaneuh:
This PR goal is to add a system to auto-correct CS and force user to fix it on PR thanks to travis.

You can find the configuration on the `.php_cs` file, feel free to comment if you want any modification.

- [x] Install and configure CS fixer
- [x] Run CS fix
- [x] Try phpunit scrict compare fixers

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,8 +19,6 @@ env:
 matrix:
   fast_finish: true
   include:
-    - php: 5.6
-      env: CS_FIXER=run
     - php: 5.4
       env: COMPOSER_FLAGS="--prefer-lowest"
     - php: 5.6
@@ -51,7 +49,7 @@ cache:
 before_install:
   - composer selfupdate
   - composer config -g github-oauth.github.com $GITHUB_OAUTH_TOKEN
-  - composer global require phpunit/phpunit fabpot/php-cs-fixer --no-update
+  - composer global require phpunit/phpunit --no-update
   - if [ "$SYMFONY_VERSION" != "" ]; then composer require "symfony/symfony:${SYMFONY_VERSION}" --no-update; fi;
   - echo "extension = mongo.so" >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini
 
@@ -60,9 +58,7 @@ install:
   - composer global update --prefer-dist --no-interaction
   - composer update --prefer-dist --no-interaction $COMPOSER_FLAGS
 
-script:
-  - if [ "$CS_FIXER" = "run" ]; then make cs_dry_run ; fi;
-  - make test
+script: make test
 
 after_script:
   - if ([ "hhvm" != "$TRAVIS_PHP_VERSION" ] && [ "nightly" != "$TRAVIS_PHP_VERSION" ]); then php ocular.phar code-coverage:upload --format=php-clover coverage.xml; fi;
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,8 @@ env:
 matrix:
   fast_finish: true
   include:
+    - php: 5.6
+      env: CS_FIXER=run
     - php: 5.4
       env: COMPOSER_FLAGS="--prefer-lowest"
     - php: 5.6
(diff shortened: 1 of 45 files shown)


.travis.yml
```diff
@@ -49,7 +51,7 @@ cache:
 before_install:
   - composer selfupdate
   - composer config -g github-oauth.github.com $GITHUB_OAUTH_TOKEN
-  - composer global require phpunit/phpunit --no-update
+  - composer global require phpunit/phpunit fabpot/php-cs-fixer --no-update
```
theofidry: we should use the `2.0@dev` version, its far faster with its cache system

Makefile
```diff
@@ -1,2 +1,8 @@
+cs:
```
theofidry: I still think this should be composer scripts rather than a makefile no?

theofidry:
Good stuff, I like it a lot although never bothered too much with that. But wouldn't be there a way to use StyleCI rather? This would be great as it would be a separated check and keep the travis for the tests only rather than doing the syntax checking

Makefile
Soullivaneuh: This is just to simplify command.

Dev want to test ? Just `make test`.

Fix cs is needed ? `make cs` instead of painful long command. ;-)

Soullivaneuh:
I prefer php-cs-fixer cli for versionned config file.

Futhermore, if we use StyleCI and want to modify configuration, we have to do it on both of them, that can produce human mistake.

It's just one job on Travis, not a big deal IMO.

Makefile
theofidry: `composer test` and `composer cs`, just 4 characters longer :dancer:

I personally don't bother too much with that as I like having everything clear in the travis, I mean the build process in it and not in a separated script. And I often roll out my own command for fast testing, usually not the same as the one on travis. But as the makefile is already here there is not much reason to move away from it

theofidry:
You can configure StyleCI with a `.styleci.yml` file so the config is versionned too right?

Soullivaneuh:
Hmm, indeed. But:

* Is StyleCI regulary updated compare to php-cs-fixer?
* Can we add custom fixer with StyleCI if needed.

Personally I would prefer Travis solution for the moment, but it's up to you to choose. ;-)

.travis.yml
Soullivaneuh: 2.0 is not stable yet. I would be glad to open a new PR when it will come. ;-)

Makefile
Soullivaneuh: Oh I didn't know we can do that. Do you have related docs?

Soullivaneuh: > But as the makefile is already here there is not much reason to move away from it

Yes, or we could replace it by composer script on another PR. :+1:

theofidry:
Looks like not for now so let's keep it that way. Just need a rebase then

Soullivaneuh:
Done.

Soullivaneuh:
@theofidry Finaly, I testes StyleCI and it works great: https://github.com/Soullivaneuh/IsoCodesValidator/pull/38

Find a quite proper solution to manage only one config file. I'll maybe propose it as a package.

If you open an account and link it on this project, I can integrate it instead of travis.

theofidry:
> If you open an account and link it on this project, I can integrate it instead of travis.

You mean enabling StyleCI?

Soullivaneuh:
> You mean enabling StyleCI?

Exactly, yes. :-)

theofidry:
https://styleci.io/repos/11691968

Soullivaneuh:
@theofidry Okay you need also to set the header comment manually on config section:

```
This file is part of the Hautelook\AliceBundle package.

(c) Baldur Rensch <brensch@gmail.com>

For the full copyright and license information, please view the LICENSE
file that was distributed with this source code.
```

This is the only part that cannot be configured on StyleCI config file.

I just asked for it here: https://twitter.com/soullivaneuh/status/644795113399582720

theofidry:
Done :)

:+1: for the tweet, also wanted to ask for that

Makefile
Soullivaneuh: @theofidry Could you provide related doc for composer script? I'll take a look and maybe submit another PR. ;-)

Soullivaneuh:
Okay StyleCI integrated. Wait for Travis.

Makefile
theofidry: pretty much:

```json
{
    "scripts": {
        "my-script-name": "MyVendor\\MyClass::postUpdate",
        "test:phpunit": "phpunit -c app",
        "test": [
            "MyVendor\\MyClass::postPackageInstall",
            "phpunit -c app"
        ]
    }
}
```

Soullivaneuh: Ok I'll take a look.

Soullivaneuh:
Tests passes! :green_heart:

theofidry:
Good work :)  :dancers:

PR #128, opened 2015-09-18
Fix typo on path for cache and log ignore

Soullivaneuh:


theofidry:
Damn, I was sure to have fixed that already

PR #136, opened 2015-09-22
Fix deprecation remove version warning

Soullivaneuh:
According to [Semantic Versioning](http://semver.org/), BC breaks should be done on new **major** version after deprecate it on a minor.

Removing this command name is a BC break, should be done on 2.0, not 1.1.

### DIFF
--- a/src/Doctrine/Command/LoadDataFixturesCommand.php
+++ b/src/Doctrine/Command/LoadDataFixturesCommand.php
@@ -148,7 +148,7 @@ class LoadDataFixturesCommand extends Command
         if (false !== strpos($input->getFirstArgument(), 'hautelook_alice:fixtures:load')
             || false !== strpos($input->getFirstArgument(), 'h:f:l')
         ) {
-            $output->writeln('<comment>The use of "hautelook_alice:fixtures:load" command is deprecated and will be removed 1.1.0. Use the
+            $output->writeln('<comment>The use of "hautelook_alice:fixtures:load" command is deprecated since 1.0 and will be removed in 2.0. Use the
 "hautelook_alice:doctrine:fixtures:load" instead.</comment>');
         }
 


src/Doctrine/Command/LoadDataFixturesCommand.php
```diff
@@ -148,7 +148,7 @@ protected function execute(InputInterface $input, OutputInterface $output)
         if (false !== strpos($input->getFirstArgument(), 'hautelook_alice:fixtures:load')
             || false !== strpos($input->getFirstArgument(), 'h:f:l')
         ) {
-            $output->writeln('<comment>The use of "hautelook_alice:fixtures:load" command is deprecated and will be removed 1.1.0. Use the
+            $output->writeln('<comment>The use of "hautelook_alice:fixtures:load" command is deprecated since 1.0 and will be removed in 2.0. Use the
```
theofidry: Well, as the command was introduced in the 1.0 *beta* I'm thinking of removing it earlier than the 2.0, what do you think?

Soullivaneuh: IMO, as beta mean "no more dev modification, just fixes", functionalities should be frozen like stable version.

It's mean to you, but I would recommend the major one.

Soullivaneuh: If users require your project with `~1.0` and use `hautelook_alice:fixtures:load` on automatic scripts for example, everything will be broken after a simple composer update.

But, as you say, it's a beta so should be used with care.

theofidry: Refering to semver it you introduce BC break changes in the beta, and we added lots of one with lots of new features in the beta. But well, I guess we can keep it as it doesn't cost much anyway.

Soullivaneuh: So, should we close this PR?

Soullivaneuh: Or not. :-)

PR #137, opened 2015-09-22
Update alice-processors.md

odoucet:
Explain another way to add Alice processor by overloading function getProcessors() in TestLoader


odoucet:
Made this PR too fast .. I see that AliceBundle does not call getProcessors() on TestLoader yet, so this pull request cannot work.

PR #145, opened 2015-10-06
Improve bare faker services declaration

Soullivaneuh:
Should be `calls`, not `call`.

http://symfony.com/doc/current/book/service_container.html#optional-dependencies-setter-injection

### DIFF
--- a/src/Resources/doc/faker-providers.md
+++ b/src/Resources/doc/faker-providers.md
@@ -57,10 +57,8 @@ services:
         lazy: true
         arguments:
             - %hautelook_alice.locale%
-        call:
-            - method: seed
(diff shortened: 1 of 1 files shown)


theofidry:
Would be great to check if I haven't done this mistake somewhere else too...

But prefer the other annotation but I guess it's a matter of personal test. Still, can you add spaces between brackets? I don't do that for PHP arrays but in YAML it's more readable this way.

Soullivaneuh:
Well, I strictly followed Symfony docs.

But yes, I can add spaces. :+1:

Soullivaneuh:
Spaces added.

PR #146, opened 2015-10-06
Fix Faker generator declarations

theofidry:
Faker generators should be generated by [`Faker\Factory::create()`](https://github.com/fzaninotto/Faker/blob/master/src/Faker/Factory.php#L17).

cc @Soullivaneuh

### DIFF
--- a/src/Resources/doc/faker-providers.md
+++ b/src/Resources/doc/faker-providers.md
@@ -54,6 +54,7 @@ instance. AliceBundle provides a faker generator `hautelook_alice.faker` configu
 services:
     hautelook_alice.bare_faker:
         class: Faker\Generator
+        factory: [ Faker\Factory, create ]
         lazy: true
         arguments:
             - %hautelook_alice.locale%


src/Resources/doc/faker-providers.md
```diff
@@ -54,13 +54,13 @@ instance. AliceBundle provides a faker generator `hautelook_alice.faker` configu
...
     hautelook_alice.bare_faker:
         class: Faker\Generator
+        factory: [ Faker\Factory, create ]
         lazy: true
         arguments:
             - %hautelook_alice.locale%
         call:
             - method: seed
               arguments: [ %hautelook_alice.seed% ]
-            - %hautelook_alice.seed%
```
Soullivaneuh: You should revert this and accept #145 instead.

src/Resources/config/services.xml
```diff
@@ -7,6 +7,7 @@
     <services>
         <service id="hautelook_alice.faker" class="Faker\Generator" lazy="true">
             <argument>%hautelook_alice.locale%</argument>
+            <factory class="Faker\Factory" method="create" />
```
Soullivaneuh: Why using a factory here? 

BTW, this is not SF 2.3 compliant.

theofidry: Yeah I've seen it my mistake, I think I've mistook the PR fail with Scrutinizer... Have to fix it

PR #147, opened 2015-10-06
Using __DIR__ on .php_cs

Soullivaneuh:
This is more reliable to do this way.


Soullivaneuh:
ping

theofidry:
My, miss that one sorry. Thanks!

PR #149, opened 2015-10-08
Show loading error messages on limit exception

### LINKED ISSUE #117: Strange LoadingLimitException case
With the following fixtures files:

```yml
AppBundle\Entity\Server:
    server (template):
        hostedDomain: sample.net
        mac: <macAddress()>
        sshUser: toor
        sshPort: 4242
        mysqlHost: localhost
        mysqlRoot: <password()>
        webServer: apache
        webServerSslListen: 0.0.0.0
        webServerSslPort: 443
        mysqlServer: true
        status: 1
        installedAt: <dateTime()>
    server_{1..9} (extends server):
        name: server_<current()>
```

I got the following error:

```bash
 [Hautelook\AliceBundle\Alice\DataFixtures\LoadingLimitException]                                                                                         
  Loading files limit of 5 reached. Could not load the following files: /home/sullivan/projects/project/src/AppBundle/DataFixtures/ORM/server.yml
```

And it works well if I remove this line:

```yml
        installedAt: <dateTime()>
```

I don't think this is the real error. I think upper exception is catched but no way to know what is happening.

Any solution for this?

Soullivaneuh:
Fixes #117.

Sample output:

```
  [Hautelook\AliceBundle\Alice\DataFixtures\LoadingLimitException]                                                                 
  Loading files limit of 5 reached. Could not load the following files:                                                            
  /home/sullivan/projects/my/project/src/AppBundle/DataFixtures/ORM/ticket.yml:                                               
   - Entity with Id @user_support_14: and Class AppBundle\Entity\User not found                                                    
   - Entity with Id @user_support_19: and Class AppBundle\Entity\User not found                                                    
   - Entity with Id @user_support_20: and Class AppBundle\Entity\User not found                                                    
   - Entity with Id @user_support_7: and Class AppBundle\Entity\User not found                                                     
  /home/sullivan/projects/my/project/src/AppBundle/DataFixtures/ORM/cron.yml:                                                 
   - Instance mask "domain_*" did not match any existing instance, make sure the object is created after its references            
  /home/sullivan/projects/my/project/src/AppBundle/DataFixtures/ORM/server.yml:                                               
   - Instance mask "user_customer_42_*" did not match any existing instance, make sure the object is created after its references  
  /home/sullivan/projects/my/project/src/AppBundle/DataFixtures/ORM/domain.yml:                                               
   - Instance mask "server_*" did not match any existing instance, make sure the object is created after its references            
```

### DIFF
--- a/src/Alice/DataFixtures/Loader.php
+++ b/src/Alice/DataFixtures/Loader.php
@@ -45,6 +45,11 @@ class Loader implements LoaderInterface
     private $loadingLimit;
 
     /**
+     * @var string[]
+     */
+    private $errorMessages;
+
+    /**
      * @param FixturesLoaderInterface $fixturesLoader
      * @param ProcessorChain          $processorChain
      * @param bool                    $persistOnce
@@ -80,6 +85,7 @@ class Loader implements LoaderInterface
         $loadFileAttempts = 0;
         $normalizedFixturesFiles = $this->normalizeFixturesFiles($fixturesFiles);
 
+        $this->errorMessages = [];
         while (true) {
             $objects = array_merge($objects, $this->tryToLoadFiles($persister, $normalizedFixturesFiles, $objects));
 
@@ -88,7 +94,7 @@ class Loader implements LoaderInterface
             }
 
             if ($this->loadingLimit <= $loadFileAttempts) {
-                throw new LoadingLimitException($this->loadingLimit, $normalizedFixturesFiles);
+                throw new LoadingLimitException($this->loadingLimit, $normalizedFixturesFiles, $this->errorMessages);
             }
 
             ++$loadFileAttempts;
@@ -194,7 +200,7 @@ class Loader implements LoaderInterface
 
                 $objects = array_merge($objects, $dataSet);
             } catch (\UnexpectedValueException $exception) {
-                // continue
+                $this->registerErrorMessage($fixtureFilePath, $exception->getMessage());
             }
         }
 
@@ -222,5 +228,24 @@ class Loader implements LoaderInterface
                 $processor->postProcess($object);
             }
         }
+    }
+
+    /**
+     * Registers the error message with the related fixture file.
+     *
+     * @param string $fixtureFilePath
+     * @param string $errorMessage
+     */
+    private function registerErrorMessage($fixtureFilePath, $errorMessage)
+    {
+        if (true === empty($errorMessage)) {
+            return;
+        }
+
+        if (!isset($this->errorMessages[$fixtureFilePath])) {
+            $this->errorMessages[$fixtureFilePath] = [];
+        }
+
+        array_push($this->errorMessages[$fixtureFilePath], $errorMessage);
     }
 }
--- a/src/Alice/DataFixtures/LoadingLimitException.php
+++ b/src/Alice/DataFixtures/LoadingLimitException.php
@@ -22,8 +22,9 @@ class LoadingLimitException extends \RuntimeException
      * @param int   $limit
      * @param array $normalizedFixturesFiles Array where keys are fixtures files path and value is a boolean set to
      *                                       true for when the fixture file has been loaded and false otherwise.
+     * @param array $errorMessages           All encountered errors messages while trying to load fixtures.
      */
-    public function __construct($limit, array $normalizedFixturesFiles)
+    public function __construct($limit, array $normalizedFixturesFiles, array $errorMessages = [])
     {
         $unloadedFiles = [];
 
@@ -33,10 +34,22 @@ class LoadingLimitException extends \RuntimeException
             }
         }
 
-        $this->message = sprintf(
-            'Loading files limit of %d reached. Could not load the following files: %s',
-            $limit,
-            implode(', ', $unloadedFiles)
-        );
+        $messageLines = [];
+        $messageLines[] = sprintf('Loading files limit of %d reached. Could not load the following files:', $limit);
+
+        sort($unloadedFiles);
+        foreach ($unloadedFiles as $unloadedFile) {
+            if (isset($errorMessages[$unloadedFile]) && 0 < count($errorMessages[$unloadedFile])) {
+                $messageLines[] = sprintf('%s:', $unloadedFile);
+                $fileErrorMessages = array_unique($errorMessages[$unloadedFile]);
+                foreach ($fileErrorMessages as $errorMessage) {
+                    $messageLines[] = sprintf(' - %s', $errorMessage);
+                }
+            } else {
+                $messageLines[] = $unloadedFile;
+            }
+        }
+
+        $this->message = implode(PHP_EOL, $messageLines);
     }
 }
--- a/tests/Alice/DataFixtures/LoaderTest.php
+++ b/tests/Alice/DataFixtures/LoaderTest.php
@@ -231,4 +231,32 @@ class LoaderTest extends \PHPUnit_Framework_TestCase
         $loader = new Loader($fixturesLoaderProphecy->reveal(), new ProcessorChain([]), false, 10);
         $loader->load($persisterProphecy->reveal(), ['random/file']);
     }
+
+    /**
+     * @covers ::load()
+     * @covers ::registerErrorMessage()
+     */
+    public function testLoaderLimitWithMessages()
+    {
+        $this->setExpectedException(
+            '\Hautelook\AliceBundle\Alice\DataFixtures\LoadingLimitException',
+            'Loading files limit of 3 reached. Could not load the following files:'.PHP_EOL
+            .'another/file:'.PHP_EOL
+            .' - That is a failed'.PHP_EOL
+            .'empty/message'.PHP_EOL
+            .'random/file:'.PHP_EOL
+            .' - Some dummy message'
+        );
+
+        $persisterProphecy = $this->prophesize('Nelmio\Alice\PersisterInterface');
+        $persisterProphecy->persist()->shouldNotBeCalled();
+
+        $fixturesLoaderProphecy = $this->prophesize('Hautelook\AliceBundle\Alice\DataFixtures\Fixtures\LoaderInterface');
+        $fixturesLoaderProphecy->load('random/file', [])->willThrow(new \UnexpectedValueException('Some dummy message'));
+        $fixturesLoaderProphecy->load('another/file', [])->willThrow(new \UnexpectedValueException('That is a failed'));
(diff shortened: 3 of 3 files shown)


src/Alice/DataFixtures/LoadingLimitException.php
```diff
@@ -33,10 +34,22 @@ public function __construct($limit, array $normalizedFixturesFiles)
...
-            'Loading files limit of %d reached. Could not load the following files: %s',
-            $limit,
-            implode(', ', $unloadedFiles)
-        );
+        $messageLines = [];
+        $messageLines[] = sprintf('Loading files limit of %d reached. Could not load the following files:', $limit);
+
+        sort($unloadedFiles);
+        foreach ($unloadedFiles as $unloadedFile) {
+            if (isset($errorMessages[$unloadedFile]) && count($errorMessages[$unloadedFile]) > 0) {
```
theofidry: use yoda conditions: `0 < count($errorMessages[$unloadedFile])`

Soullivaneuh: Yoda condition is for equal and different signs, not greater or lower.

Soullivaneuh: http://symfony.com/doc/current/contributing/code/standards.html#structure

tests/Alice/DataFixtures/LoaderTest.php
```diff
@@ -231,4 +231,28 @@ public function testLoaderWithCustomLimit()
         $loader = new Loader($fixturesLoaderProphecy->reveal(), new ProcessorChain([]), false, 10);
         $loader->load($persisterProphecy->reveal(), ['random/file']);
     }
+
```
theofidry: needs the `@cover` annotation

src/Alice/DataFixtures/Loader.php
```diff
@@ -223,4 +228,23 @@ private function persist(PersisterInterface $persister, array $objects)
...
+
+    /**
+     * Registers the error message with the related fixture file.
+     *
+     * @param string $fixtureFilePath
+     * @param string $errorMessage
+     */
+    private function registerErrorMessage($fixtureFilePath, $errorMessage)
+    {
+        if (empty($errorMessage)) {
```
theofidry: I agree it's not a big deal with `empty` and `is*, but I prefer to test it agains actual values, like here `if (true === empty($errorMessage))`. That's how it's been done in the whole library so let's stay consistant

src/Alice/DataFixtures/Loader.php
```diff
@@ -45,6 +45,11 @@ class Loader implements LoaderInterface
     private $loadingLimit;
 
     /**
+     * @var string[]
+     */
+    private $errorMessages = [];
```
theofidry: When is it actually reset? Shouldn't it be declared in the load function?

tests/Alice/DataFixtures/LoaderTest.php
Soullivaneuh: I'm not used to use this annotation cause I think it could be erroneous.

BTW, the `@cover` method is no present on another exception tests.

If you want it, what should I put on this?

tests/Alice/DataFixtures/LoaderTest.php
```diff
@@ -231,4 +231,28 @@ public function testLoaderWithCustomLimit()
...
+    public function testLoaderLimitWithMessages()
```
theofidry: As always :+1: for tests, but I would add a second call to the test just to ensure the error messages are properly reset between each loading

Soullivaneuh: I'm not sure to understand what you want exactly.

src/Alice/DataFixtures/LoadingLimitException.php
theofidry: > In programming jargon, Yoda conditions (also called Yoda notation) is a programming style where the two parts of an expression are reversed from the typical order in a conditional statement.

True yoda conditions applies for everything and I kind of find it mess up to do only for half conditions and for the sake of consistancy let's stick to the true Yoda conditions (it's like that for the whole bundle so...)

tests/Alice/DataFixtures/LoaderTest.php
theofidry: add it with the methods tested. The `@cover` is just to specify which functions are actually tested even if the test doesn't fully cover the function

theofidry:
Good one, I like it :)

tests/Alice/DataFixtures/LoaderTest.php
theofidry: What I want to ensure is that it goes as the following:

- run loading one time
- fail: get a message "message A"
- run loading one more time
- fail: get a message "message B"

And not:

- run loading one time
- fail: get a message "message A"
- run loading one more time
- fail: get a message "message A" + "message B"

src/Alice/DataFixtures/LoadingLimitException.php
Soullivaneuh: Yeah but Symfony CS does not apply full yoda condition. :-)

As discussed on another thread, it's useful to avoid mistake with assignation (`=`) but change nothing for other operators.

tests/Alice/DataFixtures/LoaderTest.php
Soullivaneuh: > add it with the methods tested.

Yeah, that's my question. Which ones?

> The `@cover` is just to specify which functions are actually tested even if the test doesn't fully cover the function

Yeah, this is exactly what I don't like... If I want, I can set `@cover toto` without test the toto method. I think it's lying IMO. :-)

tests/Alice/DataFixtures/LoaderTest.php
Soullivaneuh: Oh I see, will try.

src/Alice/DataFixtures/LoadingLimitException.php
theofidry: Oh yeah it doesn't have any practical value to force it to operators, but personally I find it inconvenient to have both conditions type at the same time.

>Yeah but Symfony CS does not apply full yoda condition. :-)

It should... we can open a ticket for that but otherwise it's not much of an issue atm.

tests/Alice/DataFixtures/LoaderTest.php
Soullivaneuh: Hmm I don't know how to do.

I tested by calling $loader->load a second time with a setExpectedException with a wrong message, tests passes...

Any idea of how to apply this test?

tests/Alice/DataFixtures/LoaderTest.php
theofidry: >Yeah, this is exactly what I don't like... If I want, I can set @cover toto without test the toto method. I think it's lying IMO. :-)

It's lying to the reader but fortunately the reader don't care because he will rely on the coverage report, for which it won't be a lie but just an information as which functions will be covered an prevent unintended coverage.

src/Alice/DataFixtures/LoadingLimitException.php
Soullivaneuh: > It should...

Maybe (I not agree), but this is the Symfony CS rules. Should we follow them or do you want this yoda condition in any case?

> we can open a ticket for that

Already discussed and refused on php-cs-fixer AFAIK.

> I find it inconvenient to have both conditions type at the same time.

Well, tastes and colors... :-)

tests/Alice/DataFixtures/LoaderTest.php
Soullivaneuh: > he will rely on the coverage report

You're telling me that `@cover` does not have influence to coverage files generation?

src/Alice/DataFixtures/LoadingLimitException.php
theofidry: >Maybe (I not agree), but this is the Symfony CS rules. Should we follow them or do you want this yoda condition in any case?

I say "it should", in the sense that the Symfony Yoda is not the yoda so doesn't make sense to me not having only the Symfony one.

> we can open a ticket for that
Already discussed and refused on php-cs-fixer AFAIK.

Then so be it, I won't waste energy for that :/

>I find it inconvenient to have both conditions type at the same time.
Well, tastes and colors... :-)

*Choosing between yoda and non-yoda conditions* is more a question of tastes and colors. Having only one is more a practical question to me. When you're dead tired I like to help your brain by providing somethings he's used to. If you have the same convention everywhere you will never think twice about it.

tests/Alice/DataFixtures/LoaderTest.php
theofidry: >You're telling me that @cover does not have influence to coverage files generation?

It does, that's the point. If I have:

```php
function testSomething()
{
    // call ::functionA() and ::functionB()
}
```

It will add coverage for both `::functionA()` and `::functionB()`. But by adding `@cover ::functionA`, the coverage for `::functionB()` will be ignored.

src/Alice/DataFixtures/LoadingLimitException.php
theofidry: From FriendsOfPHP/PHP-CS-Fixer#611 it looks like it does

Soullivaneuh: This version was removed: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/700

And I'm pretty sure they discussed about applies yoda only for == like symbols.

In any way, it's specified on the [Symfony CS standards](http://symfony.com/doc/current/contributing/code/standards.html#structure):

> this applies to ==, !=, ===, and !==

Soullivaneuh: See also https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/611#issuecomment-57337056.

src/Alice/DataFixtures/Loader.php
Soullivaneuh: I'll put it before the while start.

Soullivaneuh:
@theofidry review fixed. :+1:

theofidry:
Thank you @Soullivaneuh

Soullivaneuh:
Hey @theofidry, could we have a stable release for this? Thanks!

theofidry:
Here you go! [v1.1.0](https://github.com/hautelook/AliceBundle/releases/tag/v1.1.0). Sorry I'm not being very active right now very very busy days at work, hopefully it will get better in the coming weeks ;)

Soullivaneuh:
:+1:

PR #150, opened 2015-10-08
Prevent php-cs-fixer to apply psr0 fixer

Soullivaneuh:
Actually, if you run the fixer, it will change all class namespaces on tests folder.


theofidry:
:+1: It's been so long I haven't run the fixer thanks for pointing that out!

PR #161, opened 2015-11-05
allow this bundle to be installed with symfony3

benji07:


Soullivaneuh:
3.x is not stable yet.

:-1: for now.

BTW, if we want it, in this case it should not be on allowed_failure anymore for Travis.

benji07:
I know, but we tried to upgrade one of our projects to Symfony3, so I submit a PR to every dependency we have)

Soullivaneuh:
Then in this case you should use the `as` option of composer.

Soullivaneuh:
After some searches, I saw that SF 2.8 is [features frozen](http://symfony.com/doc/current/contributing/community/releases.html#schedule).

As 3.0 is globally the same as 2.8 but without deprecated stuff, I think we can allow it.

But in this case, tests for 2.8 and 3.0 MUST be retired from `allowed_failure` and MUST works.

Could you please take a look?

benji07:
yes, i will try to do this

benji07:
@Soullivaneuh why do you have the `theofidry/psysh-bundle` in your require-dev dependencies ?

theofidry:
> @Soullivaneuh why do you have the theofidry/psysh-bundle in your require-dev dependencies ?

Can be removed but is useful to debug from time to time, as it's a dev dependency no harm is done.

> 3.x is not stable yet.
:-1: for now.

We can still supports Sf 3.x @Soullivaneuh, I would just not *guarantee* it to work before its stable, but nothing prevents us to make it compatible already. If it implies breaking some things done may think of releasing a 2.0.

theofidry:
@benji07 does PsyshBundle causes you any trouble?

benji07:
@theofidry yes, but i can't fix it :( in the Shell class your override the renderException method and its signature change in sf 3.0 https://github.com/symfony/symfony/blob/2.8/src/Symfony/Component/Console/Application.php#L635 & https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Console/Application.php#L582

theofidry:
Hm strange by the look of [Psysh composer.json](https://github.com/bobthecow/psysh/blob/master/composer.json#L17) it should be compatible with Sf 3.0. Might be worth to open an issue for that

theofidry:
Nevermind my comment: https://github.com/bobthecow/psysh/issues/247

theofidry:
@Soullivaneuh do you think we should wait for Sf 3.0 to be released? It's good to be merge as it is IMO. Obviously there is still some thing to do to make it work but we can already include the Travis test for Sf 3.0.

Soullivaneuh:
> do you think we should wait for Sf 3.0 to be released? It's good to be merge as it is IMO.

After reflexion, I think I'll do it for my libraries too. But yes, we can't guarantee this will work.

theofidry:
@benji07 can you squash your commits ? :)

benji07:
@theofidry done

theofidry:
Thank you @benji07

PR #162, opened 2015-11-09
Use branch alias

Soullivaneuh:
Will be useful if someone want a 1.x version but with latest commits not released yet.

E.g.: `^1.0.1` or `~1.0,>1.0.0`.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -55,5 +55,10 @@
     },
     "autoload-dev": {
         "psr-4": { "Hautelook\\AliceBundle\\Tests\\": "tests" }
+    },
+    "extra": {
+        "branch-alias": {
+            "dev-master": "1.x-dev"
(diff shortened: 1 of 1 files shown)


theofidry:
He, looks like Scrutinizer has some trouble to follow up what's going on...

Soullivaneuh:
I think Scrutinizer is simply drunk. ;-)

theofidry:
As it's a composer thing it doesn't require a release does it?

Soullivaneuh:
> As it's a composer thing it doesn't require a release does it?

Exactly. As you can see, the job is already done: https://packagist.org/packages/hautelook/alice-bundle#dev-master ;-)

theofidry:
ha, indeed

PR #164, opened 2015-11-20
Fix fixtures Loader inefficiency

adrienbrault:
So after defining more than 5 fixtures file, I hit the LoadingLimitException.
I figured I had to write my own DataLoader::getFixtures to return the files
in the right order ... but this did not change anything.
After looking at the code, it seems that even with the right order, you
needed a high loading_limit configuration.

This fixes that, which should also make the fixtures process faster!


theofidry:
As fixtures are not loaded in a deterministic way and because I find it shitty for the developer to have to worry about the fixtures order. So as a result the idea is to try load the maximum number of files, keep the references of the object created, then retry again... until every files is loaded or the limit reached. This bit was indeed missing, thanks for the patch @adrienbrault.

By the way if you have other feedbacks they're obviously welcomed ;)

theofidry:
@adrienbrault as the Travis issue seems to come from PHPCR ODM and need some time to be solved, I recommend you to use a fork for now if you need the patch.

Soullivaneuh:
@adrienbrault Could you please open another PR agains `1.x` branch? Travis should be green now. :+1:

hacfi:
I just noticed this issue as well and this fix would definitely help (I fixed it by changing line 194 to `$dataSet = $this->fixturesLoader->load($fixtureFilePath, array_merge($references, $objects));`).

@theofidry Don’t fully understand why yet but in some cases the limit is reached too early.

Soullivaneuh:
> in some cases the limit is reached too early.

@hacfi Please not that you can now set `loading_limit` option on you configuration file: https://github.com/hautelook/AliceBundle#installation

theofidry:
This line is actually not needed as `$references` is updated L90 :)

hacfi:
@theofidry True. But within the `foreach` (https://github.com/hautelook/AliceBundle/blob/1.x/src/Alice/DataFixtures/Loader.php#L188) new entries are not considered: https://github.com/hautelook/AliceBundle/blob/1.x/src/Alice/DataFixtures/Loader.php#L194

I’m still in favour or merging this.

theofidry:
Ok will try to rebase that tonight

PR #166, opened 2015-11-20
Update composer.json

theofidry:


Soullivaneuh:
A little description of why you are doing this would be great. :-P

theofidry:
Ah, so old can't remember^^

PR #169, opened 2015-12-10
Use Symfony stable releases

theofidry:
Use Symfony 2.8 et 3.0 stable releases instead of the dev one.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,8 +31,8 @@ matrix:
     - php: 7.0
     - php: nightly
     - php: hhvm
-    - env: SYMFONY_VERSION=2.8.*@dev
-    - env: SYMFONY_VERSION="3.0.*@dev"
+    - env: SYMFONY_VERSION=2.8.*
+    - env: SYMFONY_VERSION=3.0.*
     # TODO: Remove those two line when test will be fixed.
     - env: COMPOSER_FLAGS="--prefer-lowest"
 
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -31,8 +31,8 @@ matrix:
     - php: 7.0
     - php: nightly
     - php: hhvm
-    - env: SYMFONY_VERSION=2.8.*@dev
-    - env: SYMFONY_VERSION="3.0.*@dev"
+    - env: SYMFONY_VERSION=2.8.*
```
Soullivaneuh: This should be simply removed from `allow_failures` and `@dev` should be removed here: https://github.com/hautelook/AliceBundle/pull/169/files#diff-354f30a63fb0907d4ad57269548329e3R27

Soullivaneuh:
BTW, we can check if php7 and hhvm pass. If yes, they can be removed from `allowed_failure`.

We can take also profit to add SF 3.1 on `allowed_failures` section.

theofidry:
Removed the dev ones.

>This should be simply removed from allow_failures

Once we ensured the compatibility that is.

>BTW, we can check if php7 and hhvm pass

Yeah that would be great.

>We can take also profit to add SF 3.1 on allowed_failures section

Agreed

Soullivaneuh:
> Once we ensured the compatibility that is.

Should be OK after https://github.com/hautelook/AliceBundle/pull/173.

Note: My PRs is agains 1.x branch. Should be the same for this PR?

theofidry:
Ah right, forgot we have two branches now. Yeah we should work on the 1.x and then port it to master.

Soullivaneuh:
1.x is green now, I let you do the rebase and open another PR. ;-)

theofidry:
Closed in favour of #188

PR #171, opened 2015-12-10
Fix deprecated Not quoting a scalar starting with `@` on fixtures

Soullivaneuh:
Ref: https://github.com/nelmio/alice/pull/290


Soullivaneuh:
Travis would be greener with this.

Soullivaneuh:
Oh, didn't see you 1.x branch. Will redo that.

PR #175, opened 2015-12-10
Fix travis badge link

theofidry:


Soullivaneuh:
Maybe `1.x` should be the default branch to avoid mistake with PR?

theofidry:
You're right let's do that

PR #187, opened 2016-02-07
Minor main README tweaks

weaverryan:
Just some minor tweaks to the README.

The only interesting one is including `doctrine/data-fixtures` in the initial install. Since the vast majority of people will use this with Doctrine, it's better to help the 99% with a small note to the 1% who are probably more advanced users anyways.

Thanks!


theofidry:
Thanks @weaverryan :)

PR #189, opened 2016-02-07
Fix StyleCI

theofidry:


theofidry:
@Soullivaneuh any idea of which fixers it comes from?

theofidry:
Disabled in the meantime

PR #194, opened 2016-02-11
Enable shard option on load fixtures command

vincentchalamon:
Allow to specify `shard` option on `hautelook_alice:doctrine:fixtures:load` command:
```
php app/console hautelook_alice:doctrine:fixtures:load --env=foo --manager=foo --shard=1
```

If related connection is instance of `PoolingShardConnection`, it will switch to correct shard database identified by requested id.

### DIFF
--- a/src/Doctrine/Command/LoadDataFixturesCommand.php
+++ b/src/Doctrine/Command/LoadDataFixturesCommand.php
@@ -13,6 +13,7 @@ namespace Hautelook\AliceBundle\Doctrine\Command;
 
 use Doctrine\Bundle\DoctrineBundle\Registry;
 use Doctrine\Common\Persistence\ManagerRegistry;
+use Doctrine\DBAL\Sharding\PoolingShardConnection;
 use Hautelook\AliceBundle\Alice\DataFixtures\Fixtures\LoaderInterface as FixturesLoaderInterface;
 use Hautelook\AliceBundle\Alice\DataFixtures\LoaderInterface;
 use Hautelook\AliceBundle\Doctrine\DataFixtures\Executor\FixturesExecutorInterface;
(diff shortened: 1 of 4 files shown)


theofidry:
Cool for the feature on my side. @Soullivaneuh any to say?

Looks like the tests are failing though :/

vincentchalamon:
Travis fixed.
@theofidry I didn't add some tests about this option on the command, some help should be appreciated :)

theofidry:
Hm, I have absolutely no idea of how we could test that :/ 

One thing we could definitely do for the next version though is moving all the logic of the command in a service (which would be more easily testable). The command would then just be a wrapper to call this service.

In the meantime I like to test as much as possible but realistically there is a few bits that are not :/

vincentchalamon:
Travis fixed.
@theofidry I didn't add some tests about this option on the command, some help should be appreciated :)

vincentchalamon:
@theofidry OK so waiting for new command structure for testing this feature ?

theofidry:
After a check there is some functional tests for this command but no unit one. This could be easily unit tested but as the unit test for this command doesn't exist I won't force you to add one for this PR to be merged (although always welcomed).

vincentchalamon:
@theofidry I just added some functional tests for this feature

theofidry:
Ha nice one!

vincentchalamon:
Seems Scrutinizer is blocked

theofidry:
Thank you @vincentchalamon :)

vincentchalamon:
Thanks @theofidry :)

theofidry:
Done ;)

vincentchalamon:
Yeah, thanks @theofidry :smile:

theofidry:
Btw if you want to give your grain of salt for the next version, check https://github.com/nelmio/alice/issues/265. Unless there is a need for it, I don't think there will be a HautelookAliceBundle v2 before alice v3.

vincentchalamon:
OK I'll check for it

PR #202, opened 2016-03-02
[WIP] Added Uploaded File provider

### LINKED ISSUE #44: Image upload Fzaninotto Faker Symfony2
To create a class Image, I have followed the steps to create a FileUpload Class in the documentary [How to Handle File Uploads with Doctrine][1]

What i am trying to do is to **use fzaninotto faker to populate the image table**. 

The image class is the following:  

	
	/**
	 * @ORM\Entity
	 * @ORM\HasLifecycleCallbacks
	 */
	class Image {
	//TODO CHECK attributbes were initially public
	    /**
	     * @ORM\Id
	     * @ORM\Column(type="integer")
	     * @ORM\GeneratedValue(strategy="AUTO")
	     */
	    private $id;
	
	    /**
	     * @ORM\Column(type="string", length=255)
	     * @Assert\NotBlank
	     */
	    private $name;
	
	    /**
	     * @ORM\Column(type="string", length=255, nullable=true)
	     */
	    private $path;
	
	    private $temp;
	
	    /**
	     * @Assert\File(maxSize="6000000")
	     */
	    private $file;
	
	    public function getAbsolutePath()
	    {
	        return null === $this->path
	            ? null
	            : $this->getUploadRootDir().'/'.$this->id.'.'.$this->path;
	    }
	
	    public function getWebPath()
	    {
	        return null === $this->path
	            ? null
	            : $this->getUploadDir().'/'.$this->path;
	    }
	
	    protected function getUploadRootDir()
	    {
	        // the absolute directory path where uploaded
	        // documents should be saved
	        return __DIR__.'/../../../../web/'.$this->getUploadDir();
	    }
	
	    protected function getUploadDir()
	    {
	        // get rid of the __DIR__ so it doesn't screw up
	        // when displaying uploaded doc/image in the view.
	        return 'uploads/images';
	    }
	
	    /**
	     * Get file.
	     *
	     * @return UploadedFile
	     */
	    public function getFile()
	    {
	        return $this->file;
	    }
	
	    /**
	     * Sets file.
	     *
	     * @param UploadedFile $file
	     */
	    public function setFile(UploadedFile $file = null)
	    {
	        $this->file = $file;
	        // check if we have an old image path
	        if (is_file($this->getAbsolutePath())) {
	            // store the old name to delete after the update
	            $this->temp = $this->getAbsolutePath();
	        } else {
	            $this->path = 'initial';
	        }
	    }
	
	    /**
	     * @ORM\PrePersist()
	     * @ORM\PreUpdate()
	     */
	    public function preUpload()
	    {
	        if (null !== $this->getFile()) {
	            $this->path = $this->getFile()->guessExtension();
	        }
	    }
	
	    /**
	     * @ORM\PostPersist()
	     * @ORM\PostUpdate()
	     */
	    public function upload()
	    {
	        if (null === $this->getFile()) {
	            return;
	        }
	
	        // check if we have an old image
	        if (isset($this->temp)) {
	            // delete the old image
	            unlink($this->temp);
	            // clear the temp image path
	            $this->temp = null;
	        }
	
	        // you must throw an exception here if the file cannot be moved
	        // so that the entity is not persisted to the database
	        // which the UploadedFile move() method does
	        $this->getFile()->move(
	            $this->getUploadRootDir(),
	            $this->id.'.'.$this->getFile()->guessExtension()
	        );
	
	        $this->setFile(null);
	    }
	
	    /**
	     * @ORM\PreRemove()
	     */
	    public function storeFilenameForRemove()
	    {
	        $this->temp = $this->getAbsolutePath();
	    }
	
	    /**
	     * @ORM\PostRemove()
	     */
	    public function removeUpload()
	    {
	        if (isset($this->temp)) {
	            unlink($this->temp);
	        }
	    }
	
	
	
	    /**
	     * Get id
	     *
	     * @return integer 
	     */
	    public function getId()
	    {
	        return $this->id;
	    }
	
	    /**
	     * Set name
	     *
	     * @param string $name
	     * @return Image
	     */
	    public function setName($name)
	    {
	        $this->name = $name;
	    
	        return $this;
	    }
	
	    /**
	     * Get name
	     *
	     * @return string 
	     */
	    public function getName()
	    {
	        return $this->name;
	    }
	
	    /**
	     * Set path
	     *
	     * @param string $path
	     * @return Image
	     */
	    public function setPath($path)
	    {
	        $this->path = $path;
	    
	        return $this;
	    }
	
	    /**
	     * Get path
	     *
	     * @return string 
	     */
	    public function getPath()
	    {
	        return $this->path;
	    }
	}

The yml file to generate the fake data is the following:

	VM\ElementBundle\Entity\Image:
	    image_{0..20}:
	        id: <current()>
	        name: image_name_<current()>
	        file: <imageUrl($width = 640, $height = 480)>

In the yml file I am not sure how to fill the other attributes, and do i have to call the 'upload' function in the class image to upload the file and generate the links ? 

  [1]: http://symfony.com/doc/current/cookbook/doctrine/file_uploads.html

Koc:
Work in progress. Closes #44

### DIFF
--- a/src/Faker/Provider/UploadedFileProvider.php
+++ b/src/Faker/Provider/UploadedFileProvider.php
@@ -1,0 +1,22 @@
+<?php
+
+namespace Hautelook\AliceBundle\Faker\Provider;
+
+use Symfony\Component\HttpFoundation\File\UploadedFile;
+
+class UploadedFileProvider
+{
+    public static function uploadedFile($filePath, $originalName = null)
+    {
+        $tmpFilePath = tempnam(sys_get_temp_dir(), 'alice-');
+        copy($filePath, $tmpFilePath);
+
+        if (null === $originalName) {
+            $originalName = basename($filePath);
+        }
+
+        //TODO: register shutdown function for removing
+
+        return new UploadedFile($tmpFilePath, $originalName, null, null, null, true);
+    }
+}
--- a/src/Resources/config/services.xml
+++ b/src/Resources/config/services.xml
@@ -20,6 +20,12 @@
                  class="Hautelook\AliceBundle\Faker\Provider\ProviderChain"
                  lazy="true" />
 
+        <service id="hautelook_alice.faker.provider_uploaded_file"
+                 class="Hautelook\AliceBundle\Faker\Provider\UploadedFileProvider"
+        >
+            <tag name="hautelook_alice.faker.provider" />
+        </service>
+
         <service id="hautelook_alice.doctrine.executor.fixtures_executor"
                  class="Hautelook\AliceBundle\Doctrine\DataFixtures\Executor\FixturesExecutor"
                  lazy="true" />


theofidry:
Hm since it's just a faker provider, couldn't it get merged back to faker itself instead?

jvasseur:
It's using the Symfony `UploadedFile` so putting it in the bundle seems logical to me.

theofidry:
Ha indeed missed it.

Fair then, but it need some tests and doc then, could you add that as well please @Koc?

Koc:
Yes, of course. Other quesion: should we add method which returns File instance (parent of UploadedFile)?

theofidry:
>Other quesion: should we add method which returns File instance (parent of UploadedFile)?

Absolutely no idea, do you have a use case?

theofidry:
Any update regarding this @Koc?

theofidry:
Closing that one for inactivity. Please feel free to open a new one on master if you have the use case for it :)

PR #204, opened 2016-03-10
Add quotes for PHP versions

theofidry:


Soullivaneuh:
Just one question: Why? :-)

theofidry:
Oh my sorry for the rogue merge.

Short story long, because otherwise version numbers are interpreted as numbers instead of strings and could be an issue in some cases; documenting it without quotes was a mistake on Travis part.

For the longer version see https://github.com/api-platform/core/pull/449#discussion_r55392343

PR #206, opened 2016-03-14
Fix fixtures Loader inefficiency

adrienbrault:
Was https://github.com/hautelook/AliceBundle/pull/164

Reopened against the `1.x` branch

### DIFF
--- a/src/Alice/DataFixtures/Loader.php
+++ b/src/Alice/DataFixtures/Loader.php
@@ -199,6 +199,7 @@ class Loader implements LoaderInterface
                 }
 
                 $objects = array_merge($objects, $dataSet);
+                $references = array_merge($references, $dataSet);
             } catch (\UnexpectedValueException $exception) {
                 $this->registerErrorMessage($fixtureFilePath, $exception->getMessage());
             }


theofidry:
Hm seems like the tests are failing, will you be able to take care of it @adrienbrault or do you wish for me to continue? (doable but it's just gonna have to wait a bit like a few days)

theofidry:
Hm, tried a little blackfire: https://blackfire.io/profiles/compare/5eb6a961-5b3a-49b8-8d0e-573896611f11/graph

Cannot really see the difference. But from what you're saying, there was a clear difference. Would be worth checking more complex scenario.

Version used for the benchmark: https://github.com/TFidryForks/AliceBundle/tree/bugfix/loader
command: `blackfire --reference=2 --samples 50 run php benchmark.php`

cc @Soullivaneuh @vincentchalamon

geoffrey-brier:
This is not only an improvement in terms of performance but also a bug fix, so huge :+1: 

The bug was quite hardcore to find so let me illustrate it. Let's say we have 3 fixtures files:

```YAML
# City.yml

App\City:
   city-paris:
       name: Paris

```

```YAML
# Project.yml

App\Project:
    project-alice:
        name: Alice
        city: @city-paris
```

```YAML
# Strength.yml

App\Strength:
    city-strength-{1..2}:
        id: <current()>
        name: Strength <current()>
        city: @city-paris
    project-strength-{1..2}:
        name: Strength Alice <current()>
        project: @project-alice
```


Thus we have a city called "Paris", a project called "Alice", Paris has 2 strengths and so does the project. Pretty simple in theory, well ... no so simple in reality.

What happens in the background:

// First call to Loader->tryToLoadFiles()

   City.yml     => Loaded correctly,
   Project.yml  => Not loaded because city-paris was not in the $references array,
   Strength.yml => Not loaded because city-paris AND project-alice were not in the $references array

// Second call to Loader->tryToLoadFiles()

   City.yml     => Already loaded,
   Project.yml  => Could be loaded because the $references contains the city,
   Strength.yml => Not loaded because project-alice were not in the $references array

And here is the problem! Though the loading failed for project-strength-1 and project-strength-2, it did not failed for city-strength-1 and city-strength-2 (eg: they were added to Paris)! But wait, there is more ...

// Third call to Loader->tryToLoadFiles()

   City.yml     => Already loaded,
   Project.yml  => Already loaded,
   Strength.yml => Could be loaded

Now the city has 4 strengths! Also, if you do not use cascade persist on your relations, doctrine will throw an exception telling you that he found new strengths objects which were not configured to cascade persist (eg: the 2 ghosts created from the second call).

Hope the explanation was clear enough and that you can merge this PR when you have time!

theofidry:
Thanks for the detailed breakdown @geoffrey-brier, indeed in that case the loader is completely unable to load the fixtures. I'm fixing that via https://github.com/hautelook/AliceBundle/pull/225 and the use case you gave is now properly loaded in 3 passes.

That was one nasty bug, thanks again for finding it. Hopefully in alice v3, those cases should no longer cause any problem at all \o/ (more efficient and no loading limit anymore).

PR #225, opened 2016-06-07
Fix loader relationship failures

theofidry:
Closes #206


egeloen:
@theofidry Thank you for this fix :)

geoffrey-brier:
@theofidry Awesome work man :+1:

src/Alice/DataFixtures/Loader.php
```diff
@@ -198,13 +197,13 @@ private function tryToLoadFiles(PersisterInterface $persister, array &$normalize
...
 
-                $objects = array_merge($objects, $dataSet);
+                $references = array_merge($references, $dataSet);
             } catch (\UnexpectedValueException $exception) {
                 $this->registerErrorMessage($fixtureFilePath, $exception->getMessage());
             }
         }
 
-        return $objects;
+        return $references;
```
egeloen: If you directly returns the references, you can remove the array merge done line 90 :)

theofidry: good shout, and as `array_merge` is slow, that's another big win

theofidry:
Here you go, a bit sad to do two releases the same day but quite happy to have solved that problem 😄

PR #226, opened 2016-06-07
Enhance Travis

theofidry:
- drop support for PHP 5.5 and hhvm
- reduce builds with coverage to PHP 5.6 with Sf 2.3 (lower deps) and PHP 5.6 with Sf 3.1
- remove memory limit for PHP
- cache `ocular.phar`
- remove PHPUnit from dependencies (relies on the build phar instead)
- remove now unused `Makefile`


Soullivaneuh:
Storing PHP binaries is a very very bad idea.

Maybe could you take a look at our? https://github.com/sonata-project/dev-kit/tree/master/project

theofidry:
Is it? Downloading them each time seems like a pain, especially as you have no way to specify which binaries you require at a project level.

theofidry:
What would be nice is a composer plugin to register binaries, but it's not going to happen soon :/

.travis.yml
```diff
@@ -39,18 +50,31 @@ cache:
     - $HOME/.composer/cache/files
 
 before_install:
-  - composer self-update
-  - if [ "$SYMFONY_VERSION" != "" ]; then composer require "symfony/symfony:${SYMFONY_VERSION}" --no-update; fi;
+  - |
+    if [ -n "$LEGACY" ]; then
+      composer global require phpunit/phpunit
```
theofidry: somehow even after that `phpunit` still refers to the container global phpunit instead of the composer one, any idea?

vincentchalamon: Maybe cause you're using `phpunit` on line 71 instead of `~/.composer/vendor/bin/phpunit`?

vincentchalamon: But travis is _supposed to_ give you a good version of phpunit according to current PHP version that you run

vincentchalamon:
You should add `bin` to gitignore

.travis.yml
theofidry: >Maybe cause you're using phpunit on line 71 instead of ~/.composer/vendor/bin/phpunit?

I though the composer global one would override the default one when doing a global install

>But travis is supposed to give you a good version of phpunit according to current PHP version that you run

Works ok most of the time, but there is two problems:

- That's definitely not the case in [PHP 5.4](https://travis-ci.org/hautelook/AliceBundle/jobs/136010842#L475) where the global PHPUnit version used is... `v1.3.1-14-gaae977c`
- [`--self-update` doesn't work](https://travis-ci.org/hautelook/AliceBundle/jobs/135996914#L172)

theofidry:
>You should add bin to gitignore

I would like to avoid people to guess which binaries to use all the time :/ As long as I can't specify them at Composer level (doable via a plugin)? I hardly see how I can do that without committing them.

The only drawback I see is that they are very large, making cloning a bit slower.

.travis.yml
vincentchalamon: > I though the composer global one would override the default one when doing a global install

I don't see in your travis file any configuration of adding `.composer/vendor/bin` directory to $PATH. Otherwise, maybe travis prefers its version instead of composer one

Soullivaneuh:
> Is it? Downloading them each time seems like a pain

Doing commit for binaries update produce more pain.

Doing a simple wget does not very hurt...

theofidry:
Downloading binaries just make it slower and when you clone the project you're not supposed to know which binaries you should use. That said I'm hoping to get what I want with https://github.com/bamarni/composer-bin-plugin, but there is still a few things to address.

Another (partial) solution would be to drop support for PHP 5.4.

Soullivaneuh:
> Downloading binaries just make it slower and when you clone the project you're not supposed to know which binaries you should use. 

See: https://travis-ci.org/sonata-project/SonataAdminBundle/jobs/136091668#L270

* PHPUnit: 0.8s
* Coveralls client: 0.02s

I hope you are joking. :wink:

Soullivaneuh:
> and when you clone the project you're not supposed to know which binaries you should use.

Specify it on README or CONTRIBUTING then.

But I strongly stay on my position: Please don't store out of project binaries on the repository.

If you really want to earn 0.8s per job, you may use the cache: https://docs.travis-ci.com/user/caching/

.travis.yml
Soullivaneuh: Downloading directly the .phar is the quickest way.

Here, a working sample: https://github.com/sonata-project/SonataAdminBundle/blob/3.2.0/.travis/install_test.sh#L6-L13

theofidry:
>PHPUnit: 0.8s
>Coveralls client: 0.02s

Ocular: 2s, that said that's a minor disagreement, not a good argument.

>Specify it on README or CONTRIBUTING then.

Same problem: nobody reads it, the problem at it's core is that there is no nice way to handle project dependencies (but maybe that plugin will solve that issue).

>But I strongly stay on my position: Please don't store out of project binaries on the repository.

Fair enough, I will wait to investigate the usage of that plugin before making my mind. 200K LoC has still an impact on the size of the project and that's for only 2 binaries.

>If you really want to earn 0.8s per job, you may use the cache: https://docs.travis-ci.com/user/caching/

Can you actually store downloaded files that way?

Soullivaneuh:
> Ocular: 2s, that said that's a minor disagreement, not a good argument.

And wrong! :stuck_out_tongue: 

https://travis-ci.org/sonata-project/SonataAdminBundle/jobs/136091668#L290

> Same problem: nobody reads it

Not our problem. If someone want to contribute on this project, the minimum to do is to read it.

> Can you actually store downloaded files that way?

Never did because never need it.

But I think you can create a bin directories and specify it on the Travis cache.

Then, use the `--timestamping` of wget command.

>   -N,  --timestamping              don't re-retrieve files unless newer than

So it would do nothing if the file already exists and is recent.

But you may have issue between phpunit version (4 and 5) if you rename them to `phpunit`.

Sincerely, doing that would earn you less than 1s per job. This totally not worth it.

greg0ire:
Don't put any binaries under version control, you will blow the size of your repository. Version control is for source code, not binaries. There is always another way.

theofidry:
>>>Ocular: 2s, that said that's a minor disagreement, not a good argument.
>>And wrong! 😛
>https://travis-ci.org/sonata-project/SonataAdminBundle/jobs/136091668#L290

Speak for yourself https://travis-ci.org/hautelook/AliceBundle/jobs/135951469#L179 xD

>Not our problem. If someone want to contribute on this project, the minimum to do is to read it.

I'm the first one annoyed by that, it puts me off to have to look for the right banaries for each of the projects when I change of computer.

>>Never did because never need it.
>But I think you can create a bin directories and specify it on the Travis cache.
>Then, use the --timestamping of wget command.
>-N, --timestamping don't re-retrieve files unless newer than
>So it would do nothing if the file already exists and is recent.

Will try that. Looks like a good thing to do regardless of the discussion of storing binaries or not. I do no see the point to add binaries (regardless the way it's added: requirement in readme or stored) that are required only for Travis.

>Don't put any binaries under version control, you will blow the size of your repository. Version control is for source code, not binaries. There is always another way.

I hear ya' guys, as I said I'm waiting on trying out https://github.com/bamarni/composer-bin-plugin which could potentially solves all of those problems before taking a decision.

Soullivaneuh:
> Speak for yourself https://travis-ci.org/hautelook/AliceBundle/jobs/135951469#L179 xD

1.1s, my last word: https://travis-ci.org/hautelook/AliceBundle/jobs/135951469#L192

:wink: 

> I hear ya' guys, as I said I'm waiting on trying out https://github.com/bamarni/composer-bin-plugin which could potentially solves all of those problems before taking a decision.

How would you implement it? Please not composer-bin-plugin is a plugin. Not all dev have this one installed.

theofidry:
>1.1s, my last word: https://travis-ci.org/hautelook/AliceBundle/jobs/135951469#L192

You are as much of a sore loser as I am 😆 On a more serious note it's the downloading that takes 1.1s but the whole step takes 1.94s no?

>How would you implement it? Please not composer-bin-plugin is a plugin. Not all dev have this one installed.

That's what I need to check, the you have a clear error when the plugin is not installed it's not much of a problem, if not it's more problematic. But otherwise I had in mind to require this plugin. A tad annoying, but it's silent project dependencies are more of a concern. And if the user doesn't want to install that plugin for a reason or another, he would still be free to download his own composer phar to install the plugin on that one only.

Soullivaneuh:
For me, in a nutshell: Use wget instead of composer for dev tools binaries.

If you want too, cache them, but this will be cumbersome comparing the earned time IMHO.

That's all, nothing more. :+1:

theofidry:
>For me, in a nutshell: Use wget instead of composer for dev tools binaries.

Why not using the plugin? That solves your project dependencies problem and some security issues as well (you no longer have to worry about checking the certificate, which is one reason why Composer is no longer recommended to be installed with wget).

Soullivaneuh:
> Why not using the plugin? That solves your project dependencies problem and some security issues as well (you no longer have to worry about checking the certificate, which is one reason why Composer is no longer recommended to be installed with wget).

I don't know this plugin. Feel free to ping me if you implemented it. :+1:

theofidry:
@Soullivaneuh RFR

PR #232, opened 2016-07-06
Adding short alias for command: fixtures:load

weaverryan:
Hi guys!

Silly as it sounds, one of the things I've never liked about this bundle is the *very* long command name - I can never remember it (and I also can't remember the short, `h:d:f:l` either).

I would just love to have this shortcut alias, instead. I know this could conflict with existing commands by this name, but will that really happen? If so, someone could send a PR in the future to make this an alias that could be turned off (but that may never happen).

If this were accepted, I'd probably update the README to use this short command.

Thanks for the consideration!

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,9 +111,9 @@ AppBundle\Entity\RelatedDummy:
         name: <name()>
 ```
 
-Then simply load your fixtures with the doctrine command `php app/console hautelook_alice:doctrine:fixtures:load` (or `php app/console h:d:f:l`).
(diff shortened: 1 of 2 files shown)


theofidry:
>(and I also can't remember the short, h:d:f:l either)

My, you're getting too old :P

Fine by me, can you update the readme as well please?

weaverryan:
@theofidry I'm just so full of wisdom, I can't remember such small details :p 

Readme updated!

theofidry:
Thanks @weaverryan :)

I'll try to do a release this weekend.

PR #174, opened 2015-12-10
Fix lowest deps

Soullivaneuh:

PR #236, opened 2016-07-11
Fix lowest deps

theofidry:
Closes #174


theofidry:
@Soullivaneuh I'm at a loss, I can't reproduce the issue and I can't see why it is failing...

PR #238, opened 2016-07-12
Fix Alice bug

theofidry:
A BC break has been introduced in alice 2.2.0, which should be fixed in 2.2.1 with https://github.com/nelmio/alice/pull/423. However this bug showed that in a specific case, we want to throw an exception and break the loading instead of trying to load the other files.


src/Alice/DataFixtures/Loader.php
```diff
@@ -199,6 +199,14 @@ private function tryToLoadFiles(PersisterInterface $persister, array &$normalize
 
                 $references = array_merge($references, $dataSet);
             } catch (\UnexpectedValueException $exception) {
+                if (1 !== preg_match(
+                        '/^Instance mask ".*" did not match any existing instance, make sure the object is created after its references$/',
```
Soullivaneuh: This should be tested. I don't think your `.*` will work correctly.

For example, I directly have this exception:

```                                   
  [UnexpectedValueException]              
  Instance user_support_4 is not defined  
```

But this does not match what you want, right?

theofidry: Indeed, will add some tests

src/Alice/DataFixtures/Loader.php
```diff
@@ -199,7 +199,20 @@ private function tryToLoadFiles(PersisterInterface $persister, array &$normalize
 
                 $references = array_merge($references, $dataSet);
             } catch (\UnexpectedValueException $exception) {
-                $this->registerErrorMessage($fixtureFilePath, $exception->getMessage());
+                $message = $exception->getMessage();
+                if (1 !== preg_match(
```
Soullivaneuh: BTW, why not using custom exception for this?

theofidry: it would be much better, but even if I were to add one now for alice, the older version still won't throw a custom exception...

PR #242, opened 2016-08-03
Load fixtures from specific directory/file à la doctrine:fixtures:load --fixtures=

fnash:


theofidry:
@fnash is there a point to add this option besides already being able to load fixture per bundle or environment?

fnash:
Absolutely. Some of my tests require DB data. Each test must have its own data set. These fixtures are stored in separate folders. So it would be nice to have this option exactly like doctrine:fixtures:load

src/Doctrine/Command/LoadDataFixturesCommand.php
```diff
@@ -185,7 +186,12 @@ protected function execute(InputInterface $input, OutputInterface $output)
 
         // Get bundles
         if (true === empty($bundles)) {
-            $bundles = $application->getKernel()->getBundles();
+            $dirOrFile = $input->getOption('fixtures');
+            if ($dirOrFile) {
+                $bundles = is_array($dirOrFile) ? $dirOrFile : array($dirOrFile);
```
theofidry: let's stick to the short array notation

src/Doctrine/Finder/FixturesFinder.php
```diff
@@ -95,6 +97,8 @@ private function getDataLoadersFromDirectory($path)
     {
         $loaders = [];
 
+        $declaredClasses = get_declared_classes();
```
theofidry: can this be put just before the foreach (L113)?

src/Finder/FixturesFinder.php
```diff
@@ -130,7 +134,7 @@ public function getFixturesFromDirectory($path)
     /**
      * Gets paths to directories containing loaders and fixtures files.
      *
-     * @param BundleInterface[] $bundles
+     * @param array $bundles
```
theofidry: `array<string, BundleInterface>`

theofidry:
Fair enough. Could you add a little functional test for it and fix the tests as well?

src/Doctrine/Finder/FixturesFinder.php
fnash: need it before all the require_once...
This reduces the foreach iterations of the reflection class.

src/Doctrine/Command/LoadDataFixturesCommand.php
```diff
@@ -185,7 +186,12 @@ protected function execute(InputInterface $input, OutputInterface $output)
 
         // Get bundles
         if (true === empty($bundles)) {
-            $bundles = $application->getKernel()->getBundles();
+            $dirOrFile = $input->getOption('fixtures');
+            if ($dirOrFile) {
+                $bundles = is_array($dirOrFile) ? $dirOrFile : [$dirOrFile];
```
fnash: @theofidry no php 5.3?

src/Doctrine/Finder/FixturesFinder.php
```diff
@@ -107,7 +111,7 @@ private function getDataLoadersFromDirectory($path)
         $loaderInterface = 'Hautelook\AliceBundle\Doctrine\DataFixtures\LoaderInterface';
 
         // Check if PHP classes are data loaders or not
-        foreach (get_declared_classes() as $className) {
+        foreach (array_diff($declaredClasses, get_declared_classes()) as $className) {
```
theofidry: can you keep the result of `array_diff($declaredClasses, get_declared_classes())`, it will be faster as well

src/Doctrine/Command/LoadDataFixturesCommand.php
theofidry: the library requires >=5.6 :)

src/Doctrine/Finder/FixturesFinder.php
theofidry: ah I see

PR #243, opened 2016-08-04
Update readMe: Basic usage

adumas37:
Updating the basic usage to use the correct syntax.


theofidry:
Thank you @adumas37 :)

PR #249, opened 2016-09-24
Upgrade to Alice 3.x

theofidry:


theofidry:
@Soullivaneuh could you check it out quickly when you have time? I would like to release a beta this weekend

PR #251, opened 2016-09-29
Allow loading fixtures from PHP files

enumag:
nelmio/alice supports [loading fixtures from plain PHP](https://github.com/nelmio/alice/blob/master/doc/complete-reference.md#php) as well which is quite useful when yaml is not enough.

### DIFF
--- a/src/Finder/FixturesFinder.php
+++ b/src/Finder/FixturesFinder.php
@@ -118,7 +118,7 @@ class FixturesFinder implements FixturesFinderInterface
     {
         $fixtures = [];
 
-        $finder = SymfonyFinder::create()->in($path)->depth(0)->files()->name('*.yml');
+        $finder = SymfonyFinder::create()->in($path)->depth(0)->files()->name('*.yml')->name('*.php');
         foreach ($finder as $file) {
             /* @var SplFileInfo $file */
             $fixtures[$file->getRealPath()] = true;


theofidry:
Indeed, could you add a little test for it please? I just want to make sure everything goes fine even if you grab loaders...

enumag:
Thanks. I'll look into it next Monday.

enumag:
@theofidry Unfortunately I'm unable to run the current tests let alone adding my own.

```
composer up
```

Failed because ext-mongo is missing. And I didn't find any way to install it on my system, only ext-mongodb.

Moving on by adding `--ignore-platform-reqs`. Composer now installed the dependencies.

Except that `phpunit/phpunit` is missing in composer.json require-dev section so I have no way to run the tests. Moving on with `composer require phpunit/phpunit --dev --ignore-platform-reqs`.

Now when I run `vendor/bin/phpunit` it fails with following error:

```
PHP Fatal error:  Cannot use 'String' as class name as it is reserved in /home/enumag/Projects/hautelook/AliceBundle/vendor/doctrine/mongodb-odm/lib/Doctrine/ODM/MongoDB/Mapping/Annotations/String.php on line 26
```

No idea how to fix that.

Can you add a guide how to run tests locally to the readme?

theofidry:
I think you can just keep going as you did and comment out the mongodb related things. It's a bit of a pain I know, this is one of the reason why it changes a lot in Alice 3.0 and AliceBundle 2.0 :/

enumag:
Well I managed to clear some mongo related things. Now it's failing when trying to connect to mysql. So I installed mysql using `sudo apt-get install mysql-server` but it's still failing - right now with `SQLSTATE[HY000] [1698] Access denied for user 'root'@'localhost'`.

theofidry:
@enumag you can see the MySQL parameters on the `phpunit.xml.dist`, I recommend you to copy it and rename it to `phpunit.xml` where you can put whatever value you want

theofidry:
if you want to you can ping me on Gitter as well for that :)

appeltaert:
What's the status on this merge? How are the tests going, need any help?

enumag:
Well mysql should be working on my localhost now but I didn't have time to go back and try AliceBundle tests again. For now my project is using the fork so it's not a priority at the moment.

Issue #209, opened 2016-03-24
I can't set seed as null


dossorio:
When I set `seed` as `null` as instructed in the docs (`seed: 1             # A seed to make sure faker generates data consistently across runs, set to null to disable`):
```yaml
hautelook_alice:
    db_drivers:
        mongodb: true
    locale: "%locale%"
    seed: null
```
I get:
```
[Symfony\Component\Config\Definition\Exception\InvalidTypeException]
  Invalid type for path "hautelook_alice.seed". Expected int, but got NULL.
  Hint: A seed to make sure faker generates data consistently across runs, set to null to disable
```

andrewjt71:
Hi, I've addressed this issue https://github.com/hautelook/AliceBundle/pull/255. Hope this helps

theofidry:
Closed via #255

PR #255, opened 2016-10-11
Allow seed to be set to null

andrewjt71:
**Related Issue**
https://github.com/hautelook/AliceBundle/issues/209

The commit https://github.com/hautelook/AliceBundle/commit/4db423f3f8c9859b6db9a8ec130c3f5bae218076 changed the node type of `seed`in the bundle configuration from a scalar node to an integer node. Although the seed value is usually an integer, integer nodes can't be set to null. [The docs](https://github.com/hautelook/AliceBundle#installation) instruct users to set the seed to null in order to disable it. This is the cause of https://github.com/hautelook/AliceBundle/issues/209.

This PR reverts this change, setting `seed` back to a scalar node (scalar nodes incorporate null - [see docs](http://symfony.com/doc/current/components/config/definition.html#node-type))


theofidry:
Thanks for the PR @andrewjt71!

If you do that then we must also add a validation for the value which was provided by the `integerNode` before, i.e. throw an `InvalidArgumentException` if the node value is not `null` and not an integer.

andrewjt71:
👍  good point. Will address this

andrewjt71:
Hi @theofidry, have added the check as requested

src/DependencyInjection/HautelookAliceExtension.php
```diff
@@ -48,6 +48,7 @@ public function prepend(ContainerBuilder $container)
      * {@inheritdoc}
      *
      * @throws InvalidConfigurationException
+     * @throws \InvalidArgumentException
```
theofidry: actually I think we can remove both throws statements, there is no point in documenting them you are not supposed to cache those kind of exceptions in the first place

src/DependencyInjection/HautelookAliceExtension.php
```diff
@@ -61,6 +62,11 @@ public function load(array $configs, ContainerBuilder $container)
         $loader = new Loader\XmlFileLoader($container, new FileLocator(__DIR__.'/../Resources/config'));
         $loader->load('services.xml');
 
+        // Ensure seed is either an integer or null.
+        if (!(is_int($config['seed']) || is_null($config['seed']))) {
+            throw new \InvalidArgumentException('Expected the seed argument to be an integer or null. Got an argument of type ' . gettype($config['seed']) . ' instead.');
```
theofidry: to be consistent with the code base, let's use a `sprintf` instead :)

andrewjt71:
Hi @theofidry, thanks for the feedback. Have addressed your points accordingly :)

theofidry:
Thanks @andrewjt71

PR #258, opened 2016-10-20
Add phpunit to dev depencies

andrewjt71:
Would be useful to add phpunit to the project's dev dependencies


theofidry:
I'm not against the idea, but if you do that you have to use this PHPUnit for testing on Travis as well.

theofidry:
Thank you @andrewjt71

PR #266, opened 2016-10-22
Declare commands as not lazy and make use of commands aliases

### LINKED ISSUE #254: The deprecated name of the command should be handled through aliases
Currently, 2 commands are registered. Instead, a single command should be registered, with the deprecated name as an alias. This way, the old name would not appear in the command listing in recent versions of Symfony (as aliases are filtered there)

theofidry:
Closes #254 and #255.


src/Doctrine/Command/LoadDataFixturesCommand.php
```diff
@@ -154,13 +154,6 @@ protected function configure()
      */
     protected function execute(InputInterface $input, OutputInterface $output)
     {
-        if (false !== strpos($input->getFirstArgument(), 'hautelook_alice:fixtures:load')
-            || false !== strpos($input->getFirstArgument(), 'h:f:l')
-        ) {
-            $output->writeln('<comment>The use of "hautelook_alice:fixtures:load" command is deprecated since 1.0 and will be removed in 2.0. Use the
```
stof: keeping the deprecation warning makes sense though

PR #288, opened 2016-12-05
fix: fixtures "disappear" when working with several bundles

kapacb:


src/Locator/EnvDirectoryLocator.php
```diff
@@ -49,7 +49,20 @@ public function locateFiles(array $bundles, string $environment): array
...
+            //
+            //        $bundle1Files = ['bundle1/001-A.php', 'bundle1/001-B.php', 'bundle1/001-C.php'];
+            //        $bundle2Files = ['bundle2/001-A.php', 'bundle2/001-B.php', 'bundle2/001-C.php', 'bundle2/001-D.php'];
+            //        ----------
+            //        var_dump($bundle1Files + $bundle2Files);
+            //        > 4 elements: 3 from bundle1 + 1 from bundle2
+            //
+            //        var_dump(array_merge($bundle1Files, $bundle2Files));
+            //        > 7 elements: 3 from bundle1 + 4 from bundle2
+            $fixtureFiles = array_merge($fixtureFiles, $this->locateBundleFiles($bundle, $environment));
```
theofidry: can you add a test for it rather than this big comment? I appreciate the explanation but a reason why I changed `array_merge` for `+` was that I thought I covered this test and `+` is faster than `array_merge`. Not sure when I changed this though

src/Locator/EnvDirectoryLocator.php
```diff
@@ -67,6 +80,10 @@ private function locateBundleFiles(BundleInterface $bundle, string $environment)
         }
 
         $files = SymfonyFinder::create()->files()->in($path)->depth(0)->name('/.*\.(ya?ml|php)$/i');
+
+        // this sort helps to set an order with filename ( "001-root-level-fixtures.yml", "002-another-level-fixtures.yml", ... )
+        $files = $files->sort( function ($a, $b) { return strcasecmp($a, $b); } );
```
theofidry: I wouldn't sort the files, it takes longer and for no reason, the ordering should not matter from AliceBundle POV. If it does for the user because some fixtures really need to be loaded first or something, this locator should be decorated to be sorted the way the user wants to

theofidry:
And congrats on making your first PR 🎉

src/Locator/EnvDirectoryLocator.php
kapacb: Is  [this](https://github.com/kapacb/AliceBundle/commit/e35236a6b1f1436ba0a663cc88126060b86e5def) enough or I need to make another PR?
( I am newbie in github :) )

Usually the number of fixture files is not great: tens or sometime hundreds(OMG) of files.
Thus performance issue is not so significant ( it is o(overall process) ).

src/Locator/EnvDirectoryLocator.php
kapacb: Once again about performance vs. usability:

- original idea was found in [Doctrine\Common\DataFixtures\OrderedFixtureInterface](https://github.com/doctrine/data-fixtures/blob/master/lib/Doctrine/Common/DataFixtures/OrderedFixtureInterface.php) (and somewhere else but now I cannot remember and find proofs)
- performance  degradation at this piece of code is not so significant but result usability grows (it becomes more "user friendly" :) )
- this can be replaced with PHP's `sort()` which is fast enough:
`-        // this sort helps to set an order with filename ( "001-root-level-fixtures.yml", "002-another-level-fixtures.yml", ... )`
`-        $files = $files->sort( function ($a, $b) { return strcasecmp($a, $b); } );`
`-`
` ...`
`-        return array_keys($fixtureFiles);`
`+       $result = array_keys($fixtureFiles);`
`+       sort($result);`
`+       return $result;`

src/Locator/EnvDirectoryLocator.php
theofidry: >Thus performance issue is not so significant ( it is o(overall process) ).

`array_merge()` is slow as it creates a new array to populate it with the arrays being merged. That said it's true that we should not care of <1000 files.

src/Locator/EnvDirectoryLocator.php
theofidry: Doctrine DataFixtures orders the fixture for a different reason, not usability but because some fixtures need to be loaded first.

I've rarely had a large set where it matters but you're right it can help and performance at this point should not matter, things like https://github.com/nelmio/alice/pull/651 are more relevant.

theofidry:
Thank you @kapacb

PR #290, opened 2016-12-13
delete cache files

allartk:
These long filenames with \ cause composer install to fail on windows. From the gitignore it seems they are commited by accident, so i assume they can be removed?


theofidry:
Indeed, thanks :)

PR #297, opened 2016-12-17
Upgrade to the last version of Alice

dunglas:


theofidry:
Ah damn forgot to check that whilst doing the releases yesterday, thanks for it

Issue #306, opened 2017-01-14
--append option throws LogicException


jverdeyen:
In the `DoctrineOrmLoader` class:

```
if ($append && $purgeWithTruncate !== null) {
            throw new \LogicException(
                'Cannot append loaded fixtures and at the same time purge the database. Choose one.'
            );
        }
```

When `$purgeWithTruncate` parameters is passed from the `DoctrineOrmLoadDataFixturesCommand` and is `=== false` at that point. Meaning using only the `--append` option will throw the `LogicException`.

This can be fixed on different location, I think it should be best to do something like this in the Command:

```
$truncate = ($input->getOption('purge-with-truncate') !== false) ? : null;
```

Thanks!

shadypierre:
I have the same issue, any news about it?

theofidry:
yes you can check #308

nifr:
PR #329 tries to fix the same issue.

theofidry:
I saw both but I think the real issue is solved in https://github.com/theofidry/AliceDataFixtures/pull/42 but I need to do a PR on HautelookAliceBundle to see if that's enough

theofidry:
This should be fixed by now :)

curry684:
So, RC1 coming out?   😉

theofidry:
Soon :) I would like a MongoDB support being merged first as it would allow to discuss some cases. Other than that I think we're good

PR #308, opened 2017-01-17
Fix --append option

jverdeyen:
Related to #306

### DIFF
--- a/tests/Console/Command/Doctrine/LoadDataFixturesCommandIntegrationTest.php
+++ b/tests/Console/Command/Doctrine/LoadDataFixturesCommandIntegrationTest.php
@@ -159,12 +159,48 @@ class LoadDataFixturesCommandIntegrationTest extends \PHPUnit_Framework_TestCase
         $this->assertFixturesDisplayEquals($expected, $commandTester->getDisplay());
     }
 
+    /**
+     * @dataProvider loadCommandProvider
+     */
+    public function testFixturesUsingAppend(array $inputs, string $expected)
+    {
+        $command = $this->application->find('hautelook:fixtures:load');
+        $commandTester = new CommandTester($command);
+        $commandTester->execute(
+            array_merge(
+                [
+                    'command' => 'hautelook:fixtures:load',
+                    '--append' => 'null',
+                ],
+                $inputs
+            ),
+            ['interactive' => false]
+        );
+
+        $this->assertFixturesDisplayEquals($expected, $commandTester->getDisplay());
+    }
+
     public function loadCommandProvider()
     {
         $data = [];
 
         $data[] = [
             [],
+            <<<'EOF'
+              > fixtures found:
+      - /home/travis/build/theofidry/AliceBundle/tests/Functional/TestBundle/Resources/fixtures/brand.yml
+      - /home/travis/build/theofidry/AliceBundle/tests/Functional/TestBundle/Resources/fixtures/product.yml
+      - /home/travis/build/theofidry/AliceBundle/tests/Functional/TestBundle/Bundle/ABundle/Resources/fixtures/aentity.php
+      - /home/travis/build/theofidry/AliceBundle/tests/Functional/TestBundle/Bundle/BBundle/Resources/fixtures/bentity.yml
+  > purging database
+  > fixtures loaded
+EOF
+        ];
+
+        $data[] = [
+            [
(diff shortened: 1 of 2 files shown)


theofidry:
Thanks for the PR @jverdeyen :)

Did you try to run it manually? I think I'm missing an integration + unit test here you didn't have to change any tests and they are still green...

jverdeyen:
Okay, I'll take a look at this. I'm currently using my patched version and it's working.

theofidry:
Need to double check with https://github.com/theofidry/AliceDataFixtures/pull/32.

Sorry for the time taken @jverdeyen, I'll have time to check it soon.

jverdeyen:
@theofidry no problem, I'm using my patched version in development for now. As fixtures are for development ;)

tests/Console/Command/Doctrine/LoadDataFixturesCommandIntegrationTest.php
```diff
@@ -178,6 +199,21 @@ public function loadCommandProvider()
...
             [
+                '--append' => null,
+            ],
+            <<<'EOF'
+              > fixtures found:
+      - /home/travis/build/theofidry/AliceBundle/tests/Functional/TestBundle/Resources/fixtures/brand.yml
+      - /home/travis/build/theofidry/AliceBundle/tests/Functional/TestBundle/Resources/fixtures/product.yml
+      - /home/travis/build/theofidry/AliceBundle/tests/Functional/TestBundle/Bundle/ABundle/Resources/fixtures/aentity.php
+      - /home/travis/build/theofidry/AliceBundle/tests/Functional/TestBundle/Bundle/BBundle/Resources/fixtures/bentity.yml
+  > purging database
```
WillGibson: It should not be saying "> purging database" when using --append. I tried taking this out locally and the test still passed, which surprised me. Hence comment I'm about to make further up re `$commandTester->getDisplay()`.

tests/Console/Command/Doctrine/LoadDataFixturesCommandIntegrationTest.php
```diff
@@ -159,6 +159,27 @@ public function testFixturesRegisteringUsingMySQL(array $inputs, string $expecte
         $this->assertFixturesDisplayEquals($expected, $commandTester->getDisplay());
     }
 
+    /**
+     * @dataProvider loadCommandProvider
+     */
+    public function testFixturesUsingAppend(array $inputs, string $expected)
```
WillGibson: This test is not needed, it's covered by `testFixturesRegisteringUsingMySQL`.

tests/Console/Command/Doctrine/LoadDataFixturesCommandIntegrationTest.php
```diff
@@ -159,6 +159,27 @@ public function testFixturesRegisteringUsingMySQL(array $inputs, string $expecte
         $this->assertFixturesDisplayEquals($expected, $commandTester->getDisplay());
```
WillGibson: @theofidry `$commandTester->getDisplay()` is returning an empty string and `$this->assertFixturesDisplayEquals` is passing regardless of whether the expected and actual match, which might be of some concern!

WillGibson:
@theofidry I have manually tested the changes here in \Hautelook\AliceBundle\Loader\DoctrineOrmLoader.php with this command...

```
./bin/console hautelook:fixtures:load -n --append --env=dev
```
...and it works fine.

Looks like there's an issue with the second build job at https://travis-ci.org/hautelook/AliceBundle/builds/192958087. I'm not sure how to fix that, but all the tests are passing for me locally.

I've made some comments on https://github.com/hautelook/AliceBundle/pull/308/files

theofidry:
Hm, to be checked with https://github.com/hautelook/AliceBundle/pull/329/files and https://github.com/theofidry/AliceDataFixtures/pull/32. All seem to relate to the same issue.

nlzet:
@theofidry Hi, is there any update on this? It seems like currently loading fixtures with the --append flag is just broken. 

Running with truncate -> OK
Running with append -> "Cannot append loaded fixtures and at the same time purge the database. Choose one." (which happens because there is a null check instead of a boolean true check)

If you would only remove the "!== null" part it is already fixed.

theofidry:
Closed by #345

PR #328, opened 2017-03-13
Make 1.x compatible with mongo ext on PHP7

omissis:
I updated the annotations on the Product Document and included Alcaues' Mongodb adapter to provide a MongoClient object when running on PHP7.


omissis:
I need to review the build on php 5 and add php 7.0 to the travis config File.

.travis.yml
```diff
@@ -1,6 +1,8 @@
 language: php
 php:
   - '5.6'
+  - '7.0'
+  - '7.1'
```
theofidry: If we are adding support for 7+ I think it makes sense to add `nightly`

.travis.yml
```diff
@@ -33,6 +35,25 @@ matrix:
         - PHPUNIT_FLAGS="--coverage-text --coverage-clover=coverage.xml"
     - php: '5.6'
       env: SYMFONY_VERSION="~3.2.0@dev"
+    - php: '7.1'
+      env:
+        - COMPOSER_FLAGS="--prefer-lowest"
+        - COVERAGE_ENABLED="true"
+        - PHPUNIT_FLAGS="--coverage-text --coverage-clover=coverage.xml"
+        - SYMFONY_VERSION="~2.3.0"
+    - php: '7.1'
```
theofidry: rather than repeating that matrix over I think you can just change `5.6` to `7.1` in the matrix above (except the prefer lowest one)

.travis.yml
```diff
@@ -33,6 +35,25 @@ matrix:
         - PHPUNIT_FLAGS="--coverage-text --coverage-clover=coverage.xml"
     - php: '5.6'
       env: SYMFONY_VERSION="~3.2.0@dev"
+    - php: '7.1'
```
theofidry: if the bridge is for PHP7 only, I would indeed add a new test with lower tests in 7+ but for `7.0` not `7.1`

.travis.yml
```diff
@@ -33,6 +35,25 @@ matrix:
...
+        - COMPOSER_FLAGS="--prefer-lowest"
+        - COVERAGE_ENABLED="true"
+        - PHPUNIT_FLAGS="--coverage-text --coverage-clover=coverage.xml"
+        - SYMFONY_VERSION="~2.3.0"
+    - php: '7.1'
+      env: SYMFONY_VERSION="~2.7.0"
+    - php: '7.1'
+      env: SYMFONY_VERSION="~2.8.0"
+    - php: '7.1'
+      env: SYMFONY_VERSION="~3.0.0"
```
theofidry: `3.0` is no longer supported, we can remove it

.travis.yml
```diff
@@ -33,6 +35,25 @@ matrix:
...
+    - php: '7.1'
+      env:
+        - SYMFONY_VERSION="~3.1.0"
+        - COVERAGE_ENABLED="true"
+        - PHPUNIT_FLAGS="--coverage-text --coverage-clover=coverage.xml"
+    - php: '7.1'
+      env: SYMFONY_VERSION="~3.2.0@dev"
```
theofidry: `3.2` is stable, so let's remove the `@dev` here and add a build for `~3.3.0@dev` :)

composer.json
```diff
@@ -18,7 +18,8 @@
     "require": {
         "php": "^5.6||^7.0",
         "nelmio/alice": "^2.1 < 2.2.0 || ^2.2.1",
-        "symfony/finder": "^2.7||^3.0"
+        "symfony/finder": "^2.7||^3.0",
+        "alcaeus/mongo-php-adapter": "^1.0"
```
theofidry: hm, not in required though

.travis.yml
omissis: +1

.travis.yml
omissis: +1

.travis.yml
omissis: +1

composer.json
omissis: I am not exactly sure how to deal with this, as the adapter is necessary for being able to use MongoClient and other deprecated classes in php7; besides, tests fail without it. Being constrained of supporting two versions of php with their incompatible mongo extensions within the 1.x branch for BC reasons, I would always require the adapter, as it skips declaring the classes if they already exist(e.g.: https://github.com/alcaeus/mongo-php-adapter/blob/master/lib/Mongo/MongoClient.php#L16). In other words: without this dependency, the code does not work on php7 and mongo due to classes missing in the php7 extension(most notably MongoClient). I agree it's not very elegant, but since I need this specific branch to work on php7 and afaik I can't conditionally require the adapter, I'd keep it. But ofc if there's a better solution, I'd be glad to implement it :)

omissis:
@theofidry I have pushed some fixes, but I need to better understand some of your comments on the matrix(I am not very familiar with it) about what to test and what not. Also note that I've added --ignore-platform-reqs to composer update to make the build work (as on php7 it relies on the adapter rather than ext-mongo)

composer.json
theofidry: Ok it seems it doesn't have much dependencies on it's own so unlikely to cause conflicts for consumers. So it's cool although not very elegant :)

theofidry:
For the matrix I mean:

- 5.6 - Symfony 2.3 - prefer-lowest
- 7.0 - Symfony 2.7 - prefer-lowest
- 7.1 - Symfony 2.7
- 7.1 - Symfony 2.8
- 7.1 - Symfony 3.1
- 7.1 - Symfony 3.2
- 7.1 - Symfony 3.3@dev
- nightly -  Symfony 3.2

theofidry:
I couldn't check your changes yet, I'll review it shortly!

omissis:
@theofidry ok, I managed to get all the dependencies right. I had to bump some doctrine's mongodb minors as the old versions were using the unsupported (by the mongodb adapter) Mongo class. Such upgrade had the unfortunate consequence of bumping up the minimum symfony version from ~2.3 to ~2.7, which caused some more conflicts. To solve such problem I bumped up all to 2.7, and all builds pass now. If you deem these changes acceptable, I'd suggest to create a new minor release in the 1.x branch, so the user of the current one won't experience conflicts on update.

theofidry:
Thanks! I'll properly look at that over the weekend and to a round of releases for alice :)

omissis:
w00t! thanks :)

omissis:
Hi @theofidry any chance you had some time to look into this? :)

theofidry:
Done :)

I'll do a release during the week/weekend :)

omissis:
@theofidry any chance you have some time for doing the release? thanks! :)

theofidry:
I'll probably get to it this week/weekend

omissis:
cool thx :)

PR #329, opened 2017-03-24
Fix append fixures

ossinkine:
`$purgeWithTruncate` variable means should database be truncated or not on purge, but not should database be purged or not.
Moreover this variable can't be null.


theofidry:
Fixed by #345

PR #336, opened 2017-04-20
Update symfony/symfony dependency.

omissis:
I have noticed I have locked the dependency to symfony/symfony a bit too strictly to 2.7 only. This PR addresses that issue allowing for 2.7 and higher.


theofidry:
Thanks

PR #338, opened 2017-05-12
FIX incorrect custom provider tag name

DaveLiddament:
The correct tag name to register new providers is actually **hautelook_alice**.faker.provider not **nelmio**.faker.provider.

Updated documentation to reflect this.


DaveLiddament:
Ah, probably reading wrong version of docs.

PR #342, opened 2017-06-13
Support fixtures outside bundles AKA Symfony Flex support

### LINKED ISSUE #296: app/Resources not scanned

dunglas:
Fix #296.

To use it with Flex:

```yml
hautelook_alice:
    fixtures_path: 'fixtures'
```

Then put your fixture files in a `fixtures/` directory at the root of the project.

### DIFF
--- a/resources/config/locator.xml
+++ b/resources/config/locator.xml
@@ -7,6 +7,7 @@
         <service id="hautelook_alice.locator.env_directory"
                  class="Hautelook\AliceBundle\Locator\EnvDirectoryLocator">
             <argument type="string">%hautelook_alice.fixtures_path%</argument>
+            <argument type="string" on-invalid="ignore">%kernel.project_dir%</argument>
         </service>
 
         <service id="hautelook_alice.locator.environmentless"
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 vendor/
 composer.lock
 coverage.xml
+fixtures/Functional/cache/
 phpunit.xml
 .php_cs.cache
 .phar/
--- a/README.md
+++ b/README.md
@@ -88,7 +88,7 @@ Configure the bundle to your needs (example with default values):
 # app/config/config_dev.yml
 
 hautelook_alice:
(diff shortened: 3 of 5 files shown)


resources/config/locator.xml
```diff
@@ -7,6 +7,7 @@
         <service id="hautelook_alice.locator.env_directory"
                  class="Hautelook\AliceBundle\Locator\EnvDirectoryLocator">
             <argument type="string">%hautelook_alice.fixtures_path%</argument>
+            <argument type="string" on-invalid="ignore">%kernel.project_dir%</argument>
```
theofidry: this is available only since 3.3 right?

dunglas: Yes but because of the `on-invalid="ignore"`, it will fallback to the previous behavior for older versions.

dunglas:
@theofidry which extra tests do you have in mind?

resources/config/locator.xml
theofidry: If possible it would be nice to support `app/Resources`, so having a `kernel.root_dir` as a fallback would be nice

theofidry:
Mostly testing it under Symfony 3.2 & 3.3 which requires Travis to be updated and the support for `app/Resources`. But maybe we can skip the later

mmarchois:
When this PR will be merged ?

theofidry:
I think running tests against Symfony 3.2 and 3.3 is a requirement for it, so Travis needs to be updated first and then the tests run again for it.

So either when someone will take care of that or when I'll have some time which won't happen before the end of the week/weekend

dunglas:
I may take a look, but do you know why Travis and Scrutinizr don't run at all? Maybe do you need to refresh the access tokens?

theofidry:
I don't think it's the issue, if you check Travis it actually ran yesterday: https://travis-ci.org/hautelook/AliceBundle/pull_requests but for some reason it takes ages and the nothing is returned

theofidry:
Meh, tests seem to be failing on Travis anyway. I'll merge this and fix that later

Issue #347, opened 2017-06-30
Locate files recursively


pdugas:
What problems would arise if the `->depth(0)` portion of the line below from [EnvDirectoryLocator.php](https://github.com/hautelook/AliceBundle/blob/master/src/Locator/EnvDirectoryLocator.php#L89) were removed?

```
$files = SymfonyFinder::create()->files()->in($path)->depth(0)->name('/.*\.(ya?ml|php)$/i');
```
I tried organizing some fixture YAML files into subdirectories today and got bit by this.

theofidry:
How is it actually a problem? I would expect it to be recursive tbh

theofidry:
Actually this may be due to [environment specific fixtures](https://github.com/hautelook/AliceBundle/blob/master/doc/advanced-usage.md#environment-specific-fixtures). I however don't see the `env` parameter in [DoctrineOrmLoadDataFixturesCommand](https://github.com/hautelook/AliceBundle/blob/master/src/Console/Command/Doctrine/DoctrineOrmLoadDataFixturesCommand.php) which is either missing or we are removing that feature from 2.x. (but I don't recall wanting to do that)

theofidry:
Closing as as explained, this is how it should work :)

PR #348, opened 2017-07-05
Update EnvDirectoryLocator.php

pdugas:
Enable recursive search for fixtures.  See #347.


theofidry:
See explanations in #347

PR #354, opened 2017-07-25
Ignore purging read only tables

zecho:
This is a hotfix to allow ignoring entities marked as read only like views etc. It also use properly the tables schema. Tested with PostgreSQL.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,6 +1,6 @@
 {
-    "name": "hautelook/alice-bundle",
-    "description": "Symfony2 Bundle to manage fixtures with Alice and Faker.",
+    "name": "zecho/alice-bundle",
+    "description": "Symfony bundle to manage fixtures with Alice and Faker.",
     "keywords": ["Symfony", "Fixture", "ORM", "Alice", "Faker"],
     "type": "symfony-bundle",
     "license": "MIT",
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -1,6 +1,6 @@
 {
-    "name": "hautelook/alice-bundle",
-    "description": "Symfony2 Bundle to manage fixtures with Alice and Faker.",
+    "name": "zecho/alice-bundle",
```
theofidry: this line should be reverted

zecho: Sorry, added to do local checks and committed it by mistake.

grigoryosifov:
+1

theofidry:
Coming back on that one:

- I feel very uncomfortable to merge anything on 1.x: the tests are broken/outdated and hard to fix due to a lot of infrastructure complexity (which has been fixed with 2.x)
- The case itself: I'm not sure why you don't want the read tables should not be deleted/truncated either
- Related tot he case above: if this is something generic and the reason is valid, I think this should be fixed in doctrine/data-fixtures instead where the purgers live

theofidry:
Closing this one as I don't think this is going to happen. I'll re-open if there is a game changer argument though :)

Issue #355, opened 2017-08-03
Fixtures not found, what I am missing?


reypm:
Hi,

I am using AliceBundle in a new project and I am following exactly the same configuration as shown [here](https://github.com/hautelook/AliceBundle#installation) meaning basically:

```
hautelook_alice:
    fixtures_path: 'Resources/fixtures/orm'
```

I have created the following files in such folder:

```
AppBundle\Entity\Post:
    post_{1..10}:
        postTitle: <text()>
        postContent: <paragraph()>
        slug: <slug()>

AppBundle\Entity\Category:
    category_1:
        categoryName: Hot
        categoryDescription: Hot topics
        slug: hot
        active: true
    category_2:
        categoryName: New
        categoryDescription: New topics
        slug: new
        active: true
```

But take a look to the following output:

```
PS C:\WINDOWS\system32> docker exec -u www-data -it reddit_webserver_1 bin/console hautelook:fixtures:load -vvv
Careful, database will be purged. Do you want to continue y/N ?y
11:23:41 INFO      [app] fixtures found
[
  "files" => []
]
[]
11:23:41 INFO      [app] fixtures loaded
[]
[]
```

What I am missing here?

theofidry:
You might be missing the environment directory, e.g. `dev` or `test`, although files at the root should be loaded as well so to me it looks like a bug.

reypm:
@theofidry I've created the folder `dev` and added the files inside it and then run the command using `--dev` option, the output is the same:

```
> docker exec -u www-data -it reddit_webserver_1 bin/console hautelook:fixtures:load --env=dev -vvv
Careful, database will be purged. Do you want to continue y/N ?y
21:52:11 INFO      [app] fixtures found
[
  "files" => []
]
[]
21:52:12 INFO      [app] fixtures loaded
[]
[]
```

theofidry:
Try to debug https://github.com/hautelook/AliceBundle/blob/master/src/Locator/EnvDirectoryLocator.php#L56 and see what's going on

SergeC:
Having the same issue. Managed to get it working with:
`# app/config/config_dev.yml`
`hautelook_alice:
    fixtures_path: '**app/**Resources/fixtures/orm' # Path to which to look for fixtures relative to the project directory or the bundle path.
`

And placed fixtures to `app/Resources/fixtures/orm/**dev/**exercise.yml`

jc-oclock:
[...]

Issue #353, opened 2017-07-21
Project root directory changes when upgrading to Symfony 3.3


nmeirik:
I noticed the following which might be relevant to mention in the documentation.

After upgrading to Symfony 3.3, the loading of fixtures started to failed without any warning. After some digging, I tracked it down to [this piece of code](https://github.com/hautelook/AliceBundle/blob/master/src/DependencyInjection/HautelookAliceExtension.php#L96) which makes use of `kernel.project_dir`if it exists (which is does in Symfony 3.3), otherwise it uses `kernel.root_dir`. I guess the latter points to the `app/`directory, while the former points to the root of the entire Symfony directory structure.

After realizing this, all I had to do was to add `app/` to the `hautelook_alice.fixtures_path` to get it working again. However, you might want to alter this behaviour (keeping the behaviour consistent across Symfony versions) or mentioning it somewhere in the documentation.

robertfausk:
Same as #355

robertfausk:
@theofidry You can close this now:)

PR #370, opened 2017-10-30
add consistent behaviour of hautelook_alice.fixtures_path

robertfausk:
across different symfony versions (<3.3 and =>3.3)
by fixing set of $rootDir in HautelookAliceExtension

This should solve #353 and #355.


theofidry:
This was added in https://github.com/hautelook/AliceBundle/pull/342 to add support for bundleless applications.

I think the real fix is to change the base dir into an array to allow to try more than one base directory. That way we can easily support both cases without any problem

robertfausk:
I assumed that kernelDir is always set and only projectDir is optional.
If you find something please let me know.

src/Locator/EnvDirectoryLocator.php
```diff
@@ -20,15 +20,18 @@
...
+    private $kernelDir;
     private $projectDir;
 
     /**
      * @param string      $fixturePath Path to which to look for fixtures relative to the bundle path.
+     * @param string      $kernelDir   Path of the kernel directory.
      * @param string|null $projectDir  Path of the project directory.
      */
-    public function __construct(string $fixturePath, string $projectDir = null)
+    public function __construct(string $fixturePath, string $kernelDir, string $projectDir = null)
```
theofidry: maybe to be more flexible it would make sense to simply pass an array of strings? That way in the extension we can set the default value to `null` (for `hautelook_alice.fixtures_path`) and if is null (as opposed to an array of string), we can use the kernel root dir and project dir .

i.e.:

```yaml
hautelook_alice:
    fixtures_path: ~ # will do as if the user passed ['%kernel.project_dir%', '%kernel.root_dir%']
                     # provided kernel.project_dir exists otherwise we omit it

    # In the both following cases we take the value defined by the user
    fixtures_path: []
    fixtures_path: ['foo']
```

theofidry:
Thanks for the work @robertfausk, I did some adjustments but the result should be the same. A noticeable change is that I got rid of the `orm` part in the default file path. I can't find a satisfying way to handle things automatically when multiple drivers are enabled but in any case this should not be handled there.

robertfausk:
Thanks for further investigation. Lgtm!

PR #374, opened 2017-11-03
Make logger argument optional in loader service

### LINKED ISSUE #373: Hard dependency on logger service
When running command in a library with barebones Symfony included:

> PHP Fatal error:  Uncaught Symfony\Component\DependencyInjection\Exception\ServiceNotFoundException: The service "hautelook_alice.loader.doctrine_orm_loader" has a dependency on a non-existent service "logger". in xxxxx/vendor/symfony/dependency-injection/Compiler/CheckExceptionOnInvalidReferenceBehaviorPass.php:31

The service has a hard unconditional dependency on the `logger` service, which is not present by default. Given the upcoming Symfony 4.0 and Flex releases which actually encourage barebones installs this will become more relevant soon.

curry684:
Fixes #373


theofidry:
Many thanks :)

PR #382, opened 2017-12-09
Fix class path in FooProvider example

bwoodmansee:
Consistent with the earlier example:

```
namespace AppBundle\DataFixtures\Faker\Provider;
```


theofidry:
Thanks @bwoodmansee. I think it will be necessary to revamp the doc a bit for Symfony4 soon but meanwhile that peculiar example is more correct :)

Issue #380, opened 2017-12-03
Feedback from using this bundle with Symfony 4.0 (Flex)


gabiudrescu:
I ran into the following problems: 
 - I had to manually declare all these bundles in config/bundles.php:
```php
    Hautelook\AliceBundle\HautelookAliceBundle::class => ['dev' => true, 'test' => true],
    Nelmio\Alice\Bridge\Symfony\NelmioAliceBundle::class => ['dev' => true, 'test' => true],
    Fidry\AliceDataFixtures\Bridge\Symfony\FidryAliceDataFixturesBundle::class => ['dev' => true, 'test' => true]
```
otherwise, flex kept complaining that FidryAliceDataFixturesBundle was not loaded - to be more specific, failed with this exception: https://github.com/hautelook/AliceBundle/blob/master/src/DependencyInjection/HautelookAliceExtension.php#L43

 - I had to change config/packages/dev/hautelook_alice.yml from:

```yml
# Read the documentation: https://github.com/hautelook/AliceBundle/blob/v1.4.1/README.md
hautelook_alice:
    db_drivers:
        orm: ~
#    locale: en_US
#    seed: 1
#    persist_once: false
#    loading_limit: 5
```

to
```yml
hautelook_alice:
    fixtures_path: 'fixtures' 
    root_dirs:
        - '%kernel.root_dir%'
        - '%kernel.project_dir%'
```

gabiudrescu:
from what I can see, only 1.4 is supported, as per https://github.com/symfony/recipes-contrib/tree/master/hautelook/alice-bundle.

are you open for a PR on supporting later versions (2.0rc)?

theofidry:
I think a flex recipe for 2.0 would be awesome.

Regarding your issue above it is a bit weird that you have to declare all the bundles if you require all of them at the same time, maybe there is more there and something could be fixed.

Regarding the config it's because there is BC breaks between 1.4 and 2.0, actually more than one :P

Czacha:
I had the same issue in Symfony 4.
After debug of EnvDirectoryLocator i delete ORM directory (because I saw that paths generated by this class doesn't have ORM at the end) and it works nice.

`
hautelook_alice:
    fixtures_path: 'DataFixtures'
`

and now I have DataFixtures dir in src folder

Nightbr:
Same problem here with Symfony Flex structure... I need to specify the path to DataFixtures...
[...]

PR #391, opened 2018-02-07
removing hard dependency for doctrine bundle

Nijusan:
As mentioned here https://github.com/hautelook/AliceBundle/issues/380#issuecomment-357746965 and in the Readme "...As FidryAliceDataFixtures is compatible with different databases/ORM, one cannot be installed by default..." the doctrine bundle shouldn't be a hard dependency when not needed.

would solve the problems for the contrib-recepie here: https://github.com/symfony/recipes-contrib/pull/252

### DIFF
--- a/src/DependencyInjection/HautelookAliceExtension.php
+++ b/src/DependencyInjection/HautelookAliceExtension.php
@@ -45,14 +45,6 @@ final class HautelookAliceExtension extends Extension
                 )
             );
         }
-        if (false === array_key_exists('Doctrine\Bundle\DoctrineBundle\DoctrineBundle', $bundles)) {
-            throw new LogicException(
-                sprintf(
-                    'Cannot register "%s" without "Doctrine\Bundle\DoctrineBundle\DoctrineBundle".',
-                    HautelookAliceBundle::class
-                )
-            );
-        }
 
         $this->loadConfig($configs, $container);
         $this->loadServices($container);


theofidry:
👍 

Could you create an empty console command which would literally just write an error message along the lines of:

```
No ORM bridge has been installed. Please install one to be able to use this command.

See https://github.com/hautelook/AliceBundle#installation for more information.
```

And then here instead of completely removing the condition check which throws a LogicException, you could register that command instead.

This way, when one install the bundle "incorrectly", he would still find the command redirecting him to some help as opposed to just be left wondering what is missing to make it work.

Nijusan:
hi @theofidry , i pushed some changes, not sure if it is what u were looking for. 
so for now, if doctrine bundle is not present, i replace the default fixture loader command with an error information command but with the same name. so when someone is calling "hautelook:fixtures:load", he will see ur error message that no orm bridge is configured.

i still think this is not perfectly right, cause when someone does not install doctrine but mongo, this check would still happen ?!

src/DependencyInjection/HautelookAliceExtension.php
```diff
@@ -45,18 +47,20 @@ public function load(array $configs, ContainerBuilder $container)
...
-                    'Cannot register "%s" without "Doctrine\Bundle\DoctrineBundle\DoctrineBundle".',
-                    HautelookAliceBundle::class
-                )
-            );
-        }
 
         $this->loadConfig($configs, $container);
         $this->loadServices($container);
 
+        if (false === array_key_exists('Doctrine\Bundle\DoctrineBundle\DoctrineBundle', $bundles)) {
```
theofidry: Could you actually replace `'Doctrine\Bundle\DoctrineBundle\DoctrineBundle'` by `\Doctrine\Bundle\DoctrineBundle\DoctrineBundle::class` and use a `use` statement to simplify the FQCN call. Even if the bundle is not installed `\Doctrine\Bundle\DoctrineBundle\DoctrineBundle::class` will resolve in the string `'Doctrine\Bundle\DoctrineBundle\DoctrineBundle'`.

To answer your question in your comment, when the MongoDB bridge will be added, via e.g. Doctrine ODM, we can add it here to ensure the "fake" command will not be registered.

Nijusan:
`composer update nothing` did not the trick for me :/

Nyholm:
I would really like to see this PR (or any variation of it)  be merged. It is currently blocking https://github.com/symfony/recipes-contrib/pull/252

welcoMattic:
Like @Nyholm says earlier, it would be really nice to merge this PR to unlock symfony/recipes-contrib#252 for everyone.

theofidry:
I'm all for merging it but it must pass the CI in order to be merged and released :)

PR #400, opened 2018-03-14
Update fixtures path

symm:
Hi,

I've just upgraded from `v2.0.0-beta.4` to `v2.0.0` and it looks like the `app/Resources/fixtures/orm/` is no longer looked at. 

Moving my fixtures from:

> src/AppBundle/Resources/fixtures/orm/Fixtures.yml 

to

> src/AppBundle/Resources/fixtures/dev/Fixtures.yml

Seems to have work, so i've updated the docs on this.


theofidry:
Thanks!

Issue #406, opened 2018-06-22
Readme faker-provider, potential bug


RomulusED69:
In this file https://github.com/hautelook/AliceBundle/blob/master/doc/faker-providers.md

```
app/config/services.yml

services:
    data_fixtures.provider.foo:
        class: AppBundle\DataFixtures\Faker\Provider\FooProvider
        tags: [ { name: nelmio_alice.faker.provider } ]
```

In SF4 if bundle is not active on prod, error because tag not exist. Maybe write a warning or write in service_test.yaml

theofidry:
Maybe the file path could be updated to match Symfony 4 when using Flex, could you do a PR for it?

RomulusED69:
Why not, but how do you want structure that? Some developers use always SF2 and 3. New files specifics to SF4 or update files with SF4 structure?

theofidry:
we could just do something like:

```yaml
# <path to the file> in Sf4
...

# <path to the file> in Sf3
...
```

And I think we can forget about Sf2.

PR #407, opened 2018-06-23
Update doc for SF4: faker-providers and alice-processors

RomulusED69:
Related to [issue](https://github.com/hautelook/AliceBundle/issues/406)


doc/alice-processors.md
```diff
@@ -4,7 +4,16 @@ Given you declared a processor `App\DataFixtures\Processor\UserProcessor`, you h
 service with the tag `fidry_alice_data_fixtures.processor` to register it:
 
 ```yaml
-# app/config/services.yml
+# config/services_dev.yml in SF4
+
+services:
+    App\DataFixtures\Processor\UserProcessor:
+        class: App\DataFixtures\Processor\UserProcessor
+        tags:
```
theofidry: to double check but I think you don't even need that tag with the autoconfigure enabled, same for faker providers, cf. https://github.com/theofidry/AliceDataFixtures/pull/65/files

RomulusED69: In my project, I tried without tag and faker doesn't work. I not tested with processor

RomulusED69:
Yes and it doesn't work

doc/alice-processors.md
greg0ire: Can you please check whether the piece of code in my PR is executed on your machine? You might have to set `autoconfigure` to `true` for that service I think

PR #409, opened 2018-06-27
Set as public hautelook_alice.fixtures.loader service

tifabien:
When we use AliceBundle with SF 3.X we get a lot of deprecation notice because the  hautelook_alice.fixtures.loader service is private.

Note that the target branch is the 1.X.


tifabien:
Hi @theofidry 
Do you think the PR could be reviewed?
Thanks a lot

theofidry:
Sorry I don't think I'll accept it. I'm fine with merging and releasing stuff on 1.x but only if the CI is green which I guess will take its load of work. If you really need this I think moving to a fork makes more sense since 1.x is no longer maintained anyway.

PR #417, opened 2018-09-18
Add PHP CS Fixer to the CI and fix the existing code

dunglas:
The config is based on [the API Platform one](https://github.com/api-platform/core/blob/master/.php_cs.dist).


tests/Alice/Loader/ServiceFactoryTest.php
```diff
@@ -32,11 +41,11 @@ public function testCanUseServiceFactory()
 
         $cityFactory = new CityFactory();
         $expected = [];
-        for ($i = 0; $i <= 10; $i++) {
+        for ($i = 0; $i <= 10; ++$i) {
```
theofidry: can this be the source of the error?

dunglas:
> Looks weird that the tests are failing now; is it rebased against master?

Yes, the problem is related to strict types. The code base isn't currently compatible with this mode.

PR #421, opened 2018-09-18
Autoconfiguration must always be possible

dunglas:
Autoconfiguration is disabled by default, and the user can always disable it for a specific service, even in file where it is enabled using `autoconfigure: true` on the given service definition.

The current check prevents of using autoconfiguration if anoter bundle registers services tagged with `nelmio_alice.faker.provider`.


theofidry:
Ah I see. Ok cool, well since this was a patch for alice 3.0 and we bumped to 3.1 this is no longer necessary (removed in #420) because it is done in nelmio/alice directly :)

Thanks for the explanation

PR #426, opened 2018-09-19
Fix Travis

dunglas:
* Fix the Travis build
* Simplify the setup

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -76,3 +76,6 @@ after_script:
     if [ -n "$COVERAGE" ]; then
       $OCULAR_BIN code-coverage:upload --format=php-clover coverage.xml
     fi
+
+notifications:
+  email: false
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,13 +30,12 @@ matrix:
     - php: '7.2'
       env:
         - COMPOSER_FLAGS="--prefer-lowest"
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -80,6 +76,3 @@ after_script:
     if [ -n "$COVERAGE" ]; then
       $OCULAR_BIN code-coverage:upload --format=php-clover coverage.xml
     fi
-
-notifications:
```
theofidry: do keep this one please, I get too many emails to be bothered by it :)

composer.json
```diff
@@ -44,8 +44,8 @@
         "doctrine/persistence": "^1.0",
         "phpunit/phpunit": "^6.0",
         "phpspec/prophecy": "^1.7",
-        "symfony/phpunit-bridge": "^3.4 || 4.0",
-        "symfony/symfony": "^3.4 || 4.0"
+        "symfony/phpunit-bridge": "^3.4 || ^4.0",
+        "symfony/symfony": "^3.4 || ^4.0"
```
theofidry: urg silly mistake, thanks

.travis.yml
dunglas: you can disable the notifications directly in the Travis UI, but as you wish.

PR #427, opened 2018-09-19
Fix the behavior when DoctrineBundle isn't installed

dunglas:
/!\\: #428 is probably a better solution

Currently when this bundle is installed but DoctrineORM isn't the whole app is broken, see https://travis-ci.org/symfony/recipes-contrib/jobs/430445745

This PR fixes this behavior.


theofidry:
Alright let's go with #428 for now then

PR #431, opened 2018-09-19
Remove dead code

dunglas:
Useless since #428


dunglas:
Failure not relates (Travis problem)

theofidry:
thanks

PR #432, opened 2018-09-19
Add traits to purge the DB and load fixtures in PHPUnit tests

dunglas:
Edit: the annotation cannot work with the current design of `KernelTestCase`, I switched to a  trait that has the other advantage of being super close of what Laravel does.

The bundle provides nice helpers, [inspired by Laravel](https://laravel.com/docs/5.6/database-testing#resetting-the-database-after-each-test), dedicated for database testing: `RefreshDatabaseTrait` and `ReloadDatabaseTrait`.
These traits allow to easily reset the database in a known state before each PHPUnit test: it purges the database then loads the fixtures.

They are particularly helpful when writing [functional tests](https://symfony.com/doc/current/testing.html#functional-tests) and when using [Panther](https://github.com/symfony/panther).

To improve performance, `RefreshDatabaseTrait` populates the database only one time, then wraps every tests in a transaction that will be rolled back at the end after its execution (regardless of if it's a success or a failure):

```php
<?php

namespace App\Tests;

use Hautelook\AliceBundle\PhpUnit\RefreshDatabaseTrait;
use Symfony\Bundle\FrameworkBundle\Test\WebTestCase;

class NewsTest extends WebTestCase
{
    use RefreshDatabaseTrait;

    public function postCommentTest()
    {
        $client = static::createClient(); // The transaction starts just after the boot of the Symfony kernel
        $crawler = $client->request('GET', '/my-news');
        $form = $crawler->filter('#post-comment')->form(['new-comment' => 'Symfony is so cool!']);
        $client->submit($form);
        // At the end of this test, the transaction will be rolled back (even if the test fails)
    }
}
```

Sometimes, wrapping tests in transactions is not possible. For instance, when using Panther, changes to the database are made by another PHP process, so it wont work.
In such cases, use the `ReloadDatabase` trait. It will purge the DB and load fixtures before every tests:

```php
<?php

namespace App\Tests;

use Hautelook\AliceBundle\PhpUnit\ReloadDatabaseTrait;
use Symfony\Component\Panther\PantherTestCase;

class NewsTest extends PantherTestCase // Be sure to extends KernelTestCase, WebTestCase or PantherTestCase
{
    use ReloadDatabaseTrait;

    public function postCommentTest()
    {
        $client = static::createPantherClient();// The database will be reset after every boot of the Symfony kernel

        $crawler = $client->request('GET', '/my-news');
        $form = $crawler->filter('#post-comment')->form(['new-comment' => 'Symfony is so cool!']);
        $client->submit($form);
    }
}
```

This strategy doesn't work when using Panther, because the changes to the database are done by another process, outside
of the transaction.

Both traits provide several configuration options as protected static properties:

* `self::$manager`: The name of the Doctrine manager to use
* `self::$bundles`: The list of bundles where to look for fixtures
* `self::$append`: Append fixtures instead of purging
* `self::$purgeWithTruncate`: Use TRUNCATE to purge
* `self::$shard`: The name of the Doctrine shard to use
* `self::$connection`: The name of the Doctrine connection to use

Use them in the `setUpBeforeClass` method.

```php
<?php

namespace App\Tests;

use Hautelook\AliceBundle\PhpUnit\RefreshDatabaseTrait;
use Symfony\Bundle\FrameworkBundle\Test\WebTestCase;

class NewsTest extends WebTestCase
{
    use RefreshDatabaseTrait;

    public static function setUpBeforeClass()
    {
        self::$append = true;
    }

    // ...
}
```

Old comment (previous commits):

---

This PR introduces a new PHPUnit listener purge the DB and load the fixtures before every test.
It can work by just purging, or to purge only one time then wrap tests in transactions that will be rollbacked at the end (faster, but not 100% reliable, and don't work for E2E tests like Panther).

Like the Symfony PHPUnit Bridge, it relies on special PHPUnit groups:

* `@group refresh-database`: purge and load the fixtures
* `@group refresh-database-transaction`: same and use transactions

This feature is heavily inspirited from [Laravel database testing traits](https://laravel.com/docs/5.6/database-testing#resetting-the-database-after-each-test) (from a DX point of view, obviously internals cannot be similar).

Example usage with Panther:

```xml
<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
    <!-- ... -->
    <listeners>
        <!-- ... -->
        <listener class="Hautelook\AliceBundle\PhpUnit\RefreshDatabaseListener" />
    </listeners>
</phpunit>
```

```yaml
App\Entity\Comment:
    comment_{1..10}:
        news: <randomElement(["week-601", "symfony-live-usa-2018"])>
        body: <paragraph()>
```

```php
<?php

namespace App\Tests;

use Symfony\Component\Panther\PantherTestCase;

/**
 * @group refresh-database
 */
class CommentsTest extends PantherTestCase
{
    public function testComments()
    {
        $client = static::createPantherClient();
        $crawler = $client->request('GET', '/news/symfony-live-usa-2018');

        $client->waitFor('#post-comment'); // Wait for the form to appear, it may take some time because it's done in JS
        $form = $crawler->filter('#post-comment')->form(['new-comment' => 'Symfony is so cool!']);
        $client->submit($form);

        $client->waitFor('#comments li:nth-child(6)'); // Wait for the comment to appear
        $this->assertSame('Symfony is so cool!', $crawler->filter('#comments li:last-child')->text());
    }

    public function testRefreshDatabase()
    {
        $client = static::createPantherClient();
        $crawler = $client->request('GET', '/news/symfony-live-usa-2018');

        $client->waitFor('#comments'); // Wait for the comments to appear

        // Be sure that the database has been reseted
        $this->assertNotSame('Symfony is so cool!', $crawler->filter('#comments li:last-child')->text());
    }
}
```

TODO:

* [x] Add tests


src/PhpUnit/RefreshDatabaseListener.php
```diff
@@ -0,0 +1,127 @@
...
+ * @experimental
+ *
+ * @author Kévin Dunglas <dunglas@gmail.com>
+ */
+final class RefreshDatabaseListener implements TestListener
+{
+    use TestListenerDefaultImplementation;
+
+    // These public properties provide a way to globally configure the listener during bootstrap
+    public static $manager = null;
```
theofidry: While I love with the "omit the types when obvious" policy to reduce the noise, the types here are not obvious so I would add them for this case

src/PhpUnit/RefreshDatabaseListener.php
```diff
@@ -0,0 +1,127 @@
...
+
+        if (null === self::$connections) {
+            self::$connections = new \SplObjectStorage();
+        }
+        self::$connections[$test] = $doctrine->getConnection(self::$connectionName);
+        self::$connections[$test]->setAutoCommit(false);
+        self::$connections[$test]->beginTransaction();
+    }
+
+    public function endTest(Test $test, $time): void
```
theofidry: can be `KernelTestCase` here. If you cannot change the signature because of the parent class, I would add an `inheritdoc` block to indicate you are implementing/overriding the parent method + add `@param`. Same for the others

src/PhpUnit/RefreshDatabaseListener.php
```diff
@@ -0,0 +1,127 @@
...
+    public function startTest(Test $test): void
+    {
+        if (!$test instanceof KernelTestCase || !\in_array('refresh-database', $groups = $test->getGroups(), true)) {
+            return;
+        }
+
+        $kernel = $this->getKernel($test);
+        $container = $kernel->getContainer();
+        $doctrine = $container->get('doctrine');
+        $transaction = \in_array('transaction', $groups, true);
```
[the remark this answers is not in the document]
dunglas: We should update the PHP CS Fixer config then

src/PhpUnit/RefreshDatabaseListener.php
dunglas: Not on this one because the listener will call it for every test. I’ll update the private methods.

src/PhpUnit/RefreshDatabaseListener.php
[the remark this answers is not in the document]
dunglas: Actually, PHP CS Fixer automatically remove import of functions to optimize it...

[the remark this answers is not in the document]
theofidry: But IIRC that's a config entry. I would love a PHP-CS-Fixer rule to do the import function statements but that's not the case for now. So in the meantime I would just disable that cofnig for PHP-CS-Fixer, Hautelook doesn't really have critical performance code anyway, if optims can be done it would rather be at Alice & the ORM level

README.md
```diff
@@ -152,6 +156,87 @@ If you want to load the fixtures of a bundle only, do `php bin/console hautelook
...
+use Hautelook\AliceBundle\PhpUnit\RefreshDatabaseTrait;
+use Symfony\Bundle\FrameworkBundle\Test\WebTestCase;
+
+class MyTest extends WebTestCase
+{
+    use RefreshDatabaseTrait;
+
+    public static function setUpBeforeClass()
+    {
+        self::$useTransactions = true; // Enable the transactional mode
```
theofidry: shouldn't that mode be the default one?

README.md
```diff
@@ -152,6 +156,87 @@ If you want to load the fixtures of a bundle only, do `php bin/console hautelook
...
+    }
+
+    public function aTest()
+    {
+        $client = static::createPantherClient();
+        // The transaction starts here
```
theofidry: how come it doesn't start earlier? I thought it would start as soon as you inter in the test

README.md
```diff
@@ -152,6 +156,87 @@ If you want to load the fixtures of a bundle only, do `php bin/console hautelook
...
+        $crawler = $client->request('GET', '/my-news');
+        $form = $crawler->filter('#post-comment')->form(['new-comment' => 'Symfony is so cool!']);
+        $client->submit($form);
+        // At the end of this test, the transaction will be rolled back
```
theofidry: Might be worth mentioning that it will be the case even though the test will fail

theofidry:
Looks like there is a test failing possible due to using a too old version of Symfony but we can bump the requirement if necessary

README.md
dunglas: It's not 100% safe, but I'll try something else.

README.md
dunglas: It starts when the kernel boot. Here it's `createPantherClient` that boots the kernel. It's intended I think. I'll reword to make this statement bolder.

README.md
theofidry: When is it not?

dunglas:
PR ready for another review!
Now there is 2 traits, one using transactions (`RefreshDatabase`), and another doing a purge before every tests for Panther (`ReloadDatabase`).

I also updated the docs.

README.md
```diff
@@ -152,6 +156,103 @@ If you want to load the fixtures of a bundle only, do `php bin/console hautelook
 Next chapter: [Advanced usage](doc/advanced-usage.md)
 
 
+## Database testing
+
+The bundle provides nice helpers, [inspirited by Laravel](https://laravel.com/docs/5.6/database-testing#resetting-the-database-after-each-test),
```
weaverryan: inspired by Laravel

README.md
```diff
@@ -152,6 +156,103 @@ If you want to load the fixtures of a bundle only, do `php bin/console hautelook
...
+dedicated for database testing: `RefreshDatabaseTrait` and `ReloadDatabaseTrait`.
```
weaverryan: > for testing with a database:

README.md
```diff
@@ -152,6 +156,103 @@ If you want to load the fixtures of a bundle only, do `php bin/console hautelook
...
+These traits allow to easily reset the database in a known state before each PHPUnit test: it purges the database then loads
```
weaverryan: > These traits allow you to easily reset the database to a known state before each PHPUnit test: it purges the database then loads

README.md
```diff
@@ -152,6 +156,103 @@ If you want to load the fixtures of a bundle only, do `php bin/console hautelook
...
+the fixtures.
+
+They are particularly helpful when writing [functional test](https://symfony.com/doc/current/testing.html#functional-tests)
```
weaverryan: functional test**s**

README.md
```diff
@@ -152,6 +156,103 @@ If you want to load the fixtures of a bundle only, do `php bin/console hautelook
...
+and when using [Panther](https://github.com/symfony/panther).
+
+To improve performances, `RefreshTrait` populates the database only one time, then wraps every tests in a
```
weaverryan: performance~s~

RefreshDatabaseTrait

README.md
```diff
@@ -152,6 +156,103 @@ If you want to load the fixtures of a bundle only, do `php bin/console hautelook
...
+<?php
+
+namespace App\Tests;
+
+use Hautelook\AliceBundle\PhpUnit\RefreshDatabaseTrait;
+use Symfony\Component\Panther\PantherTestCase;
+
+class NewsTest extends PantherTestCase // Be sure to extends KernelTestCase, WebTestCase or PantherTestCase
+{
+    use RefreshDatabaseTrait;
```
weaverryan: Should be Reload

README.md
```diff
@@ -152,6 +156,103 @@ If you want to load the fixtures of a bundle only, do `php bin/console hautelook
...
+
+    public function postCommentTest()
+    {
+        $client = static::createPantherClient(); // The database will be reseted after every boots of the Symfony kernel
```
weaverryan: // The database will be reset after every boot of the Symfony kernel

README.md
```diff
@@ -152,6 +156,103 @@ If you want to load the fixtures of a bundle only, do `php bin/console hautelook
...
+        $crawler = $client->request('GET', '/my-news');
+        $form = $crawler->filter('#post-comment')->form(['new-comment' => 'Symfony is so cool!']);
+        $client->submit($form);
+    }
+}
+```
+
+This strategy doesn't work when using Panther, because the changes to the databases are done by another process, outside
```
weaverryan: database~s~

src/PhpUnit/BaseDatabaseTrait.php
```diff
@@ -0,0 +1,72 @@
...
+use LogicException;
+use Symfony\Bundle\FrameworkBundle\Console\Application;
+use Symfony\Bundle\FrameworkBundle\Test\KernelTestCase;
+
+trait BaseDatabaseTrait
+{
+    /**
+     * @var string|null The name of the Doctrine manager to use
+     */
+    public static $manager;
```
weaverryan: Do these need to be public?

dunglas: Not anymore, it was a "rest" of the previous implementation. Updated, thanks.

dunglas:
Thanks for proof-reading @weaverryan! Updated.

theofidry:
All good for me, thanks!

README.md
```diff
@@ -152,6 +156,104 @@ If you want to load the fixtures of a bundle only, do `php bin/console hautelook
...
+
+        $crawler = $client->request('GET', '/my-news');
+        $form = $crawler->filter('#post-comment')->form(['new-comment' => 'Symfony is so cool!']);
+        $client->submit($form);
+    }
+}
+```
+
+This strategy doesn't work when using Panther, because the changes to the database are done by another process, outside
+of the transaction.
```
Oliboy50: @dunglas  since you made some references to `Panther` in this snippet, I think you want to remove these lines (to me, it looks like a duplicate of the previous snippet comment => L. 195 to 197)

PR #435, opened 2018-09-20
Fix shortcut value for manager option

benr77:
Change the short version of the `--manager` command option from `-em` which does not work, to `-m` which does.


theofidry:
thanks

PR #438, opened 2018-10-02
Remove the theofidry/alice-bundle-extension installation suggestion

tifabien:
As theofidry/alice-bundle-extension is not supported with the new v2 version of the bundle, we can remove the composer suggest section


theofidry:
Good point, thanks!

PR #439, opened 2018-10-10
Make loaded fixture available in test after they are loaded by alice.

jamesmoey:


theofidry:
Thank you @jamesmoey!

PR #444, opened 2018-11-21
Update faker-providers.md

alexandre-mace:
added Faker to simple provider namespace, otherwise there was an issue with the declared namespace in services.yml


alexandre-mace:
Ok then we need to change the declared namespace in services.yml
 `class: AppBundle\DataFixtures\Faker\Provider\FooProvider` 
into
class: `AppBundle\DataFixtures\Provider\FooProvider`

otherwise the namespace declared in the class doesnt correspond to the one in services.yml

snebes:
@theofidry This corrects the invalid documentation on the [1.x branch.](https://github.com/hautelook/AliceBundle/blob/1.x/src/Resources/doc/faker-providers.md)

theofidry:
Ah I see, thanks

PR #447, opened 2018-12-03
Add a trait to recreate the database schema

GaryPEGEOT:
Add a trait to (re)create the database schema to allow SQLite database usage.


bupyc9:
When will it be released?

PR #448, opened 2018-12-03
Fix root node deprecation in symfony/config > 4.1

mdeboer:
Fixes the deprecation warnings about tree builders without root nodes being deprecated since 4.2. This fix is BC with 4.1 and older and is similar to the [fix applied to doctrine/DoctrineBundle](https://github.com/doctrine/DoctrineBundle/pull/853/files).


theofidry:
Thanks for the patch! The CI seem to fail on something unrelated, it looks like PHP-CS-Fixer is being run against the Symfony fixture app cache which it shouldn't

PR #458, opened 2019-02-04
Update Symfony decorator configuration

snebes:
This configuration update is available for currently supported Symfony versions
https://symfony.com/doc/4.2/service_container/service_decoration.html

I did not notice a set voice or tone for your config, so feel free to update this.


theofidry:
This syntax is nothing new, it's been there since at least Symfony 2.6. So I think we can just replace the existing example with yours instead :)

snebes:
@theofidry You're right, updated the PR to reflect this.

theofidry:
Thanks!

PR #467, opened 2019-04-07
Allow multiple fixture paths during bundle configuration.

wRLSS:
hautelook#404
Allow multiple `fixture_path` entries during bundle configuration.
Keeping BC-Break with previous Configuration.

PR #470, opened 2019-05-21
Improved docs about multiple fixtures paths

jmsche:
Hi, just a little commit to improve docs for the `fixtures_path` configuration option that was updated bt #467 :)


theofidry:
Thanks!

PR #473, opened 2019-08-02
Remove final from DoctrineOrmLoadDataFixturesCommand

### LINKED ISSUE #460: The DoctrineOrmLoadDataFixturesCommand is not extendable
Hello,

It is currently impossible to override the `hautelook_alice.console.command.doctrine.doctrine_orm_load_data_fixtures_command` service (command) as the `DoctrineOrmLoadDataFixturesCommand` is currently declared as a `final class`.
In fact it's possible to override the service but we have to copy past the most of this class code which is quite frustrating.

Did I missed something ? Is it a special choice or an oversight (in which case I can do the PR) ?

Thank you for the bundle and the support you are providing.

tifabien:
Fixes #460


theofidry:
Thanks!

PR #487, opened 2019-12-04
Enable support for Symfony 5

olix21:
Work in progress. I'm doing this Pull Request to keep track of my changes and to open it for discussion


olix21:
Also depending on https://github.com/nelmio/alice/pull/1008

theofidry:
Tagged

fr4ngus:
I have made a PR [https://github.com/olix21/AliceBundle/pull/1](https://github.com/olix21/AliceBundle/pull/1) if it can help to move forward.

composer.json
```diff
@@ -32,19 +32,20 @@
...
         "doctrine/data-fixtures": "^1.2",
         "doctrine/orm": "^2.5.11",
         "psr/log": "^1.0",
-        "symfony/finder": "^3.4 || ^4.0",
-        "symfony/framework-bundle": "^3.4.24 || ^4.0",
-        "theofidry/alice-data-fixtures": "^1.0"
+        "symfony/finder": "^3.4 || ^4.0 || ^5.0",
+        "symfony/framework-bundle": "^3.4.24 || ^4.0 || ^5.0",
+        "theofidry/alice-data-fixtures": "^1.0",
+        "symfony/string": "^5.0"
```
jonag: Won't this dependency prevent users using Flex with the setting `"extra.symfony.require": "4.4.*"` to install the bundle? And we don't seem to use any feature from this component, are we?

theofidry: As far as I can tell we are not using it at all in AliceBundle so it should be possible to remove it

composer.json
```diff
@@ -32,19 +32,20 @@
...
     },
     "require-dev": {
         "doctrine/persistence": "^1.0",
         "phpunit/phpunit": "^8.5",
         "phpspec/prophecy": "^1.7",
-        "symfony/phpunit-bridge": "^3.4.31 || ^4.0"
+        "symfony/phpunit-bridge": "^3.4 || ^4.0 || ^5.0"
```
juntereiner: It should be `"symfony/phpunit-bridge": "^3.4.31 || ^4.0 || ^5.0" ` otherwise travis will fail with prefer-lowest composer flag.

fixtures/HttpKernel/DummyKernel.php
```diff
@@ -98,7 +98,7 @@ public function getBundle($name, $first = true)
     /**
      * {@inheritdoc}
      */
-    public function locateResource($name, $dir = null, $first = true)
+    public function locateResource($name)
```
juntereiner: It won't work with symfony 3.4 composer flag.

composer.json
olix21: symfony string removed

fixtures/HttpKernel/DummyKernel.php
olix21: Indeed. But if we don't change it, it will not be compatible with SF 5.
Should I handle it in the `__call` method?

composer.json
olix21:
```
1) Hautelook\AliceBundle\DependencyInjection\HautelookAliceBundleTest::testServiceRegistration
Symfony\Component\DependencyInjection\Exception\RuntimeException: You cannot use the "slugger" service since the String component is not installed. Try running "composer require symfony/string".
```

juntereiner:
There is an another issue with the last test in https://github.com/hautelook/AliceBundle/blob/master/tests/DependencyInjection/HautelookAliceBundleTest.php for SF5 compatibility.

Apparently, symfony/string is mandatory when booting the kernel in SF5.

fixtures/HttpKernel/DummyKernel.php
dunglas: Maybe could we stop supporting Symfony 3.4? Supporting only the latest LTS will make maintenance way easier.

jonag:
It may be an issue with Symfony itself : The framework bundle [registers the slugger service](https://github.com/symfony/symfony/blob/e60a876201b5b306d0c81a24d9a3db997192079c/src/Symfony/Bundle/FrameworkBundle/Resources/config/services.xml#L126) and then [throws an exception](https://github.com/symfony/symfony/blob/be0e04b84aa6dc262d45982b731b711a56f3d1a1/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php#L206) if the component isn't installed. If the component is mandatory it should be added to the composer.json of symfony/framework-bundle.

(I don't know how the framework-bundle works, I've guessed that by looking at the sources, I may by completely wrong ...)

fixtures/HttpKernel/DummyKernel.php
juntereiner: Yes, it would be probably best if we release a major version supporting only symfony 4.4 and 5.x.

theofidry: A minor version bump is fine for those things

theofidry:
I think this should be reported to Symfony itself. Meanwhile we can require it: it may mess up some some (not all) flex installs for a while, but I think it's ok. It can be fixed later in a patch release once fixed in Symfony

jonag:
According to the [PR that introduced the service](https://github.com/symfony/symfony/pull/33768) the error should only be triggered when we use the service.

I think it may have something to do with what [we are doing here](https://github.com/hautelook/AliceBundle/blob/master/fixtures/Functional/AppKernel.php#L61). Maybe iterating over the definitions or setting it to public instantiates the service and triggers the error? I don't know enough about the DI container to be sure or to be able to propose a fix...

theofidry:
That could be indeed... Can you try if skipping it would be enough?

jonag:
I'm sorry, I'm not sure to understand what you mean by skipping it?

theofidry:
Excluding it from the loop setting all the services public. But alternatively we can:

- Go on with https://github.com/hautelook/AliceBundle/pull/487#issuecomment-566199040 for now and attempt to fix it later
- Migrate toward using the test container which doesn't inline the services hence should remove the need to setting all the services to public

jonag:
I've sent a PR to @olix21 to skip the Slugger when setting the services to true. It's pretty ugly but it should fix the tests temporary.
The best option should probably to
> Migrate toward using the test container which doesn't inline the services hence should remove the need to setting all the services to public

but I'm afraid I don't have the knowledge to do that easily...

fr4ngus:
I've sent a PR to @olix21 too for the second alternative
> Migrate toward using the test container which doesn't inline the services hence should remove the need to setting all the services to public

olix21:
I did the merge. In my initial PR, I skipped 1 test because of Symfony deprecations. 
If we agree to do drop the support of old SF versions, we could remove that test.

theofidry:
I think the prefer lowest could be solved by that and the last build is failing due to CS as far as I can tell

fr4ngus:
Yep I agree with you @theofidry, I think we can drop the support of old SF too.

michaelKaefer:
Is there any way to use this bundle in Symfony 5 before this PR gets merged?

theofidry:
@michaelKaefer by forking it... But if you have time for that you might as well take over this PR :)

fr4ngus:
I've sent a PR to @olix21 and just fix CS

fr4ngus:
I've sent another PR to @olix21 and fix travis build

theofidry:
Thank you @fr4ngus and @olix21!

src/DependencyInjection/HautelookAliceExtension.php
```diff
@@ -62,12 +63,17 @@ public function load(array $configs, ContainerBuilder $container)
...
      * @throws \InvalidArgumentException
+     * @throws \Exception
      */
     private function loadConfig(array $configs, ContainerBuilder $container)
     {
         $configuration = new Configuration();
         $processedConfiguration = $this->processConfiguration($configuration, $configs);
 
+        if (Kernel::VERSION_ID >= 40000) {
+            $processedConfiguration['root_dirs'] = ['%kernel.project_dir%'];
```
bendavies: this needs to be reverted.
This is over riding any config provided by the user.

For example, in tests I have
```
hautelook_alice:
  fixtures_path: 'Resources/fixtures'
  root_dirs:
    - '%kernel.root_dir%/../tests/'
```
and this change overwrites my config.

fr4ngus: Yes, I will make another pull request to fix this.
In my opinion, we must change the default root_dirs, that by default use "%root_dir%" param that is deprecated since Symfony 4.2, and removed since 5.0.

fr4ngus:
I've made a PR #494 to fix this

PR #488, opened 2019-12-13
use pcov instead of phpdbg

juntereiner:
Update travis to use pcov for code coverage.

To keep compatibility with php7.1 and phpunit 7, pcov-clobber is used https://github.com/krakjoe/pcov-clobber


.travis.yml
```diff
@@ -47,8 +47,10 @@ before_install:
   - if [ "$SYMFONY_REQUIRE" != "" ]; then composer global require --no-scripts symfony/flex; fi
   - |
     if [ -n "$COVERAGE" ]; then
+      pecl install pcov;
+      composer require --dev pcov/clobber;
```
theofidry: since packages are installed later I would add a `--no-update`flag here otherwise it's going to end up doing the resolution twice. That said might be possible to upgrade to PHPUnit 8 to not require that package?

theofidry:
I think we can drop PHP 7.1 to upgrade to PHPUnit 8

juntereiner:
I had to play with the version constraints for travis to pass with "--prefer-lowest" flag and phpunit 8.5.

The minimum version for the Symfony/Framework is 3.4.23: https://github.com/symfony/symfony/pull/30084
For the symfony/phpunit-bridge, it is 3.4.31: https://github.com/symfony/symfony/pull/32933

theofidry:
Thanks a lot @juntereiner!

PR #493, opened 2020-01-11
Address deprecations from persistence lib

greg0ire:
A backwards-compatibility layer has been added to persistence to help
consumers move to the new namespacing. It is based on class aliases,
which means the type declaration changes should not be a BC-break: types
are the same.
See doctrine/persistence#71

This means:
- using the new namespaces
- adding autoload calls for new types to types that may be extended and
use persistence types in type declarations of non-constructor methods,
so that signature compatibility is recognized by old versions of php.
Luckily, none of those were found in this package.
More details on this at
https://dev.to/greg0ire/how-to-deprecate-a-type-in-php-48cf


greg0ire:
I don't understand the Composer failure for the prefer-lowest build :thinking:

greg0ire:
I understand now, I added it to `require` when there was already a requirement in the `require-dev` section. This looks wrong to me, by the way, since there are references to the persistence lib in `src`. Should I move it to the `require` section. Or is it there because it is considered a dependency only when using an optional feature? Since there is already a requirement for `doctrine/orm` this looks unlikely.

theofidry:
No indeed, not sure why this one is in require-dev

greg0ire:
Ok, let me move it.

greg0ire:
@theofidry PTAL

theofidry:
Thank you @greg0ire

PR #494, opened 2020-01-17
Fix overriding custom path fixtures root dir

fr4ngus:
Symfony 5.0 Support - fix issue custom root_dirs fixtures caused by PR #487


src/DependencyInjection/HautelookAliceExtension.php
```diff
@@ -47,11 +46,7 @@ public function load(array $configs, ContainerBuilder $container)
         }
 
         if ($missingBundles) {
-            throw new LogicException(sprintf(
-                'To register "%s", you also need: "%s".',
-                HautelookAliceBundle::class,
-                implode('", "', array_keys($missingBundles))
-            ));
+            throw new LogicException(sprintf('To register "%s", you also need: "%s".', HautelookAliceBundle::class, implode('", "', array_keys($missingBundles))));
```
theofidry: if that change is not needed I would prefer it to be reverted. I know that form is preferred in Symfony, but I find the current form more readable

theofidry:
Thank you @fr4ngus !

PR #496, opened 2020-01-20
Added --no-bundles flag command and `noBundles` param to DoctrineOrmLoader::load method

### LINKED ISSUE #495: Impossible to disable including fixtures from bundles with `hautelook:fixtures:load` command
My fixtures places in `tests/Fixtures` directory and I have problem with other bundles, like `nelmio/cors-bundle`, because it also has `tests/Fixtures` with `ProviderMock.php` class, which triggers error. (`ProviderMock.php` has different from implemented interface return type and php 7.2 triggers error)

Of course, I can hack this problem by adding only 1 random bundle to bundles list, but since I use symfony-flex, there is no need for me in loading fixtures from bundles, so it would be cool if there was some flag like `--no-bundles`.

A can prepare merge-request if it is ok for you.

Metalrex100:
Closes #495


Metalrex100:
@theofidry seems like tests not passes because of triggered deprecations. Do you have suggestions how could it be fixed?

src/Console/Command/Doctrine/DoctrineOrmLoadDataFixturesCommand.php
```diff
@@ -125,16 +131,31 @@ protected function execute(InputInterface $input, OutputInterface $output)
             }
         }
 
+        $noBundles = $input->getParameterOption('--no-bundles') ?? true;
```
theofidry: Let's do it another way:

- when `--no-bundles` is _not_ provided, we trigger the deprecation notice with the message:

>The configuration parameter `hautelook.root_dirs` should be used to specify the directories to include. If done or if you do not need to load bundle's fixtures, use the `--no-bundle` option

- when `--bundle` is provided, we trigger the deprecation notice with the message:

>The option `--bundle` is deprecated. Use the configuration parameter `hautelook.root_dirs` to include the desired directories instead

- when `--bundle` and `--no-bundles` are _both_ provided, throw `DomainException`

>Cannot the option `--bundle` with `--no-bundles`

theofidry: Can you also add a:

```php
TODO: make $noBundles the default in 3.x. Accept the option nonetheless to ease the upgrade but add a deprecation notice to say this option is no longer necessary
```

theofidry:
Regarding the CI failure you would need to update this number from 0 to the desired one: https://github.com/hautelook/AliceBundle/blob/master/phpunit.xml.dist#L13

tests/Console/Command/Doctrine/LoadDataFixturesCommandTest.php
```diff
@@ -139,4 +139,48 @@ public function testCallCommandWithArguments()
...
+        /** @var ManagerRegistry|ObjectProphecy $managerRegistryProphecy */
+        $managerRegistryProphecy = $this->prophesize(ManagerRegistry::class);
+        $managerRegistryProphecy->getManager('DummyManager')->willReturn($manager = new FakeEntityManager());
+        /** @var ManagerRegistry $managerRegistry */
+        $managerRegistry = $managerRegistryProphecy->reveal();
+
+        /** @var LoaderInterface|ObjectProphecy $loaderProphecy */
+        $loaderProphecy = $this->prophesize(LoaderInterface::class);
+        $loaderProphecy
+            ->load($application, $manager, ['ABundle', 'BBundle'], 'dummy_env', true, true, 'shard_id', true)
```
theofidry: shouldn't it be called with `[]` here? instead of `['ABundle', 'BBundle']`

Metalrex100: I've created new test to check if exception will be thrown if both `--bundle` and `--no-bundles` flags were specified.

I also increased max deprecations count to 50, but still tests not passing despite the fact that all assertions are ok. Do you know what is the problem?)

Metalrex100: Oh, seems like travis.yml overrides `SYMFONY_DEPRECATIONS_HELPER` env, so probably I need to update it also in travis.yml

.travis.yml
```diff
@@ -29,7 +29,7 @@ matrix:
     - php: '7.2'
       env:
         - COMPOSER_FLAGS="--prefer-lowest"
-        - SYMFONY_DEPRECATIONS_HELPER="weak_vendors"
+        - SYMFONY_DEPRECATIONS_HELPER="max[self]=50"
```
theofidry: might be worth to be removed completely

tests/Console/Command/Doctrine/LoadDataFixturesCommandTest.php
theofidry: Oh I see I misread the test case

tests/Console/Command/Doctrine/LoadDataFixturesCommandTest.php
```diff
@@ -139,4 +140,48 @@ public function testCallCommandWithArguments()
...
+            ->shouldBeCalledTimes(0);
+
+        /** @var LoaderInterface $loader */
+        $loader = $loaderProphecy->reveal();
+
+        $command = new DoctrineOrmLoadDataFixturesCommand('dummy', $managerRegistry, $loader);
+        $command->setApplication($application);
+
+        $this->expectException(DomainException::class);
```
theofidry: hm rather than relying on PHPUnit magic here, i tend to prefer:

```php
try {
  // do the thing that fails

  $this->fail(); // guard to make sure the thing above really failed
catch (DomainException $exception) {
  $this->assertSame(
    'error message',
    $exception->getMessage()
  );
}
```

theofidry:
Ha indeed it needs to be `@trigger_error()` not `trigger_error()` :)

theofidry:
Hm maybe you can try to restore the `max[self=0]` and mark the two command test cases with `@group legacy`. It's not ideal but will do the job

theofidry:
Thank you @Metalrex100!

rogamoore:
Just a quick question: The deprecation message says "The configuration parameter hautelook.root_dirs ...".
Shouldn't the key be "hautelook_**alice**.root_dirs"?

theofidry:
Indeed

PR #499, opened 2020-02-18
Use correct configuration key in deprecation message

rogamoore:
Change hautelook.root_dirs to hautelook_alice.root_dirs


theofidry:
Thanks!

PR #500, opened 2020-02-19
Updated factory documentation

benndt:
Using __constructor for a factory is deprecated since 3.0.0 and removed in 4.0.0:

https://github.com/nelmio/alice/blob/master/doc/complete-reference.md#using-a-factory--a-named-constructor


theofidry:
Thank you @benndt

PR #501, opened 2020-02-26
Display loaded fixtures files while executing the Command

juuuuuu:
So we could have the same kind of output as DoctrineFixturesBundle.

If you agree, I will add some tests.


theofidry:
I'm fine with that but not with `SymfonyStyle` in the loader: the loader itself is not console specific. It could however make use of a logger for that

juuuuuu:
> I'm fine with that but not with `SymfonyStyle` in the loader: the loader itself is not console specific. 

For my defense, I would like to remember that the loader already got a `Symfony\Bundle\FrameworkBundle\Console\Application` argument, which is pretty "console specific" :smile:.

> It could however make use of a logger for that.

I'll take a look to do this in that direction.

> This extra output could also be displayed only with a verbose mode for example

Great idea!

theofidry:
That's true, it should be replaced by the `Kernel` instead, but that's for the next major version I guess

PR #502, opened 2020-03-15
Remove `final` keyword from DoctrineOrmLoader

nirix:
Removes the `final` keyword from `DoctrineOrmLoader`, allowing it to be extended.


theofidry:
Thanks @nirix

PR #504, opened 2020-04-20
Update faker-providers.md

l-vo:
Extending Faker\Provider\Base is needed to allow autoconfigure to work


theofidry:
Indeed, thanks!

PR #517, opened 2020-09-23
Add lazy loading

pwlb:
https://symfony.com/doc/current/console/commands_as_services.html#lazy-loading
To use other commands without the external services.


theofidry:
Thank you @pwlb

PR #518, opened 2020-10-02
Allow doctrine/persistence 2.0

franmomu:
The necessary changes were made in https://github.com/hautelook/AliceBundle/pull/493.

Fixes https://github.com/hautelook/AliceBundle/issues/513, https://github.com/hautelook/AliceBundle/issues/514 and https://github.com/hautelook/AliceBundle/issues/519

~EDIT: It needs some more love 😬~

PR #520, opened 2020-10-09
Bump PHP to 7.3

franmomu:
I had to update php-cs-fixer version as the current one wasn't supporting PHP > 7.2


franmomu:
Now it's ok, I reverted bumping `theofidry/alice-data-fixtures` to 1.3 because that should be done in https://github.com/hautelook/AliceBundle/pull/518.

theofidry:
Thank you @franmomu

PR #533, opened 2021-02-21
Switch to GitHub actions

soyuka:


theofidry:
Thanks @soyuka; I think GA is not running atm because it needs to have a workflow in first so it should run from now on

VincentLanglet:
@theofidry Can you tag a new release with php8 support ? :)

